PR #3, opened 2019-08-19
ChaCha20Poly1305 AEAD

tarcieri:
Implements the ChaCha20Poly1305 AEAD described in RFC 8439 using the `chacha20` and `poly1305` crates.


tarcieri:
Removed the draft PR status on this as it's functionally complete and passing tests, but leaving `[WIP]` on since it depends on https://github.com/RustCrypto/traits/pull/40 being merged first.

cc @jcape

Cargo.toml
```diff
@@ -2,3 +2,6 @@
 members = [
     "chacha20poly1305",
 ]
+
+[patch.crates-io]
+aead = { git = "https://github.com/mobilecoinofficial/traits" }
```
tarcieri: TODO: use a release of the `aead` crate

chacha20poly1305/Cargo.toml
```diff
@@ -4,10 +4,17 @@ version = "0.0.0"
...
+description = """
+ChaCha20Poly1305 Authenticated Encryption with Additional Data Algorithm (RFC 8439)
+"""
 documentation = "chacha20poly1305"
 repository = "https://github.com/RustCrypto/AEADs"
 keywords = ["crypto", "cipher", "aead"]
 categories = ["cryptography", "no-std"]
 
 [dependencies]
+aead = "*"
```
tarcieri: TODO: change to `aead = "0.1"`

Cargo.toml
```diff
@@ -2,3 +2,6 @@
 members = [
     "chacha20poly1305",
 ]
+
+[patch.crates-io]
+aead = { git = "https://github.com/RustCrypto/traits" }
```
tarcieri: @newpavlov I just removed `[WIP]` as this PR is functionally complete. Not sure how you feel about merging this with ^^^, but otherwise I'd call this PR ready for review

tarcieri:
Note: I'd like to do XChaCha20Poly1305 as well, which is a simple extension of this, but I figured I'd start with this as a first pass.

chacha20poly1305/Cargo.toml
```diff
@@ -4,10 +4,16 @@ version = "0.0.0"
...
 documentation = "chacha20poly1305"
 repository = "https://github.com/RustCrypto/AEADs"
 keywords = ["crypto", "cipher", "aead"]
 categories = ["cryptography", "no-std"]
 
 [dependencies]
+aead = "0.1"
+chacha20 = { version = "0.2.1", features = ["zeroize"] }
+poly1305 = "0.2"
+zeroize = { version = "0.9", default-features = false }
```
newpavlov: I thought in other crates `zeroize` was optional?

chacha20poly1305/src/lib.rs
```diff
@@ -0,0 +1,157 @@
...
+
+impl NewAead for ChaCha20Poly1305 {
+    type KeySize = U32;
+
+    fn new(key: GenericArray<u8, U32>) -> Self {
+        ChaCha20Poly1305 { key }
+    }
+}
+
+impl Aead for ChaCha20Poly1305 {
```
newpavlov: You probably should implement `StatelessAead` instead. You'll get `Aead` via the [blanket impl](https://github.com/RustCrypto/traits/blob/master/aead/src/lib.rs#L87).

chacha20poly1305/src/lib.rs
```diff
@@ -0,0 +1,157 @@
...
+    fn drop(&mut self) {
+        self.key.as_mut_slice().zeroize();
+    }
+}
+
+/// ChaCha20Poly1305 instantiated with a particular nonce
+struct CipherInstance {
+    chacha20: ChaCha20,
+    poly1305: Poly1305,
+}
```
newpavlov: Looks like it can be made generic over `StreamCipher` trait? I wonder if it should not go to the `poly1305` crate...

Cargo.toml
newpavlov: I think it would be better to depend on git for `aead` directly in the `chacha20poly1305` crate. But overall I don't have any objections against temporarily using a git dependency. :)

newpavlov:
BTW I wonder if we should not modify the AEAD trait constructors to `new(key, nonce)`.

jcape:
> BTW I wonder if we should not modify the AEAD trait constructors to `new(key, nonce)`.

That would break things like STREAM that need to pass specific nonce values, wouldn't it?

Cargo.toml
tarcieri: Cargo doesn't allow you to patch dependencies at the individual crate level within a workspace. I imagine that precludes a workspace-wide dependency resolution.

chacha20poly1305/Cargo.toml
tarcieri: Happy to keep it that way if you want. The other crates have a lower MSRV than `zeroize`, so it had to be optional there. Could always be optional but on-by-default as well.

chacha20poly1305/src/lib.rs
tarcieri: Aah, my bad, I was worried about the `&mut self` that everything `salsa20-core`-based requires, but forgetting the fact I instantiate the cipher for every nonce. Will fix.

tarcieri:
@jcape correct.

Particularly for AES-GCM(-SIV) it'd be nice to do key expansion once for a particular key, and reuse that across nonces, especially when encrypting several AEAD messages in a row as in STREAM (or a transport encryption protocol like TLS or Noise).

It helps less for Salsa20 family ciphers, where you do want to buffer/cache the initial block state (https://github.com/RustCrypto/stream-ciphers/issues/50), but particularly for ChaCha20 much of the complexity of computing it comes from the nonce.

chacha20poly1305/src/lib.rs
tarcieri: This is a first cut 😉But yeah, good point.

It would definitely be nice to reuse the same generic "Poly1305 plus a Salsa20 family stream cipher" core to implement Salsa20Poly1305, ChaCha20Poly1305, and XChaCha20Poly1305.

The `poly1305` crate might make sense as it's the eponymous algorithm's primary use case. I could also see this further justifying the existence of `salsa20-core` versus trying to abstract it away into a more generic `ctr`.

Cargo.toml
newpavlov: I meant not patching, but using `aead = { git = "https://github.com/RustCrypto/traits" }` instead of `aead = "0.1"` in the `[dependencies]` section.

chacha20poly1305/Cargo.toml
newpavlov: Hm, I guess making it optional and enabling it by default will not make substantial difference compared to the non-optional approach, so we can leave it as-is.

BTW I wonder how zeroize will work in case like this:
```rust
struct Foo { .. }

struct Bar { foo: Foo, baz: Baz }
```

if both `Foo` and `Bar` implement `Zeroize`. Will it zeroize `Foo` space twice?

tarcieri: @newpavlov I've been using explicit `Drop` handlers, and generally trying to push those down to the "leaf" data owner. If there are two `Drop` handlers like in your example, they will zero `foo` twice. I'd suggest only `Foo` impl `Drop` in the example above.

tarcieri:
@newpavlov aside from making the implementation generic, look good for a first cut?

newpavlov:
Yeap, you can merge it at will.

tarcieri:
Alright! Now I guess we just gotta figure out what to do about AAD in the `aead` crate 😉

newpavlov:
BTW another idea which I had is to introduce a trait for AEAD based on stream ciphers, it would allow us to simplify detached interface. maybe we could even add an associated constant of type `enum AdPlace { Beginning, End }`, which will be used to assemble message. But I guess it will be better to discuss it in a dedicated issue/PR. :)

tarcieri:
@newpavlov yeah definitely, almost every AEAD is based on a stream cipher, so it'd be nice to simplify the common case for those

tarcieri:
Opened an issue to discuss stream cipher-based AEADs (#45)

XinGOfCloude18 (approved):

PR #4, opened 2019-08-25
XChaCha20Poly1305

tarcieri:
Adds support for XChaCha20Poly1305 AEAD: an extended 192-bit (24-byte) nonce variant of ChaCha20Poly1305 as described in:

https://tools.ietf.org/html/draft-arciszewski-xchacha-03

This variant is also implemented in libsodium and used by e.g. WireGuard.


chacha20poly1305/src/cipher.rs
```diff
@@ -0,0 +1,122 @@
+//! Core AEAD cipher implementation for (X)ChaCha20Poly1305.
+
+// TODO(tarcieri): make this reusable for (X)Salsa20Poly1305
```
tarcieri: Arguably the best way to do this is making it generic to all stream ciphers: https://github.com/RustCrypto/traits/issues/45

PR #5, opened 2019-08-29
chacha20poly1305: Use 0.1 release of `aead` crate

tarcieri:
This updates `chacha20poly1305` to switch to the newly released `aead` crate.


tarcieri:
@newpavlov can you grant access to the `chacha20poly1305` crate so I can publish it? https://crates.io/crates/chacha20poly1305

PR #11, opened 2019-09-19
AES-GCM-SIV

tarcieri:
Implementation of [AES-GCM-SIV] ([RFC 8452]) based off the pseudocode in the RFC / prose description.

[AES-GCM-SIV]: https://en.wikipedia.org/wiki/AES-GCM-SIV
[RFC 8452]: https://tools.ietf.org/html/rfc8452


tarcieri:
Removing draft (but not WIP, still some work to do)

It appears to be working 🎉 

There is a much more extensive set of test vectors I plan on adding (i.e. all of the ones from the RFC). Right now this is just doing "happy path" testing of a single test vector for Aes128GcmSiv.

This is also using an incredibly naive implementation. There's lots of small things that could get fixed up, which I may or may not do prior to merging.

tarcieri:
I've added all of the test vectors from RFC 8452 Appendix C, as well as a test for MAC verification failure, and everything is passing.

I might see if there are any low-handing fruit performance improvements I can do. AES-CTR isn't parallelized right now, and decryption doesn't interleave AES-CTR and POLYVAL.

I suppose I can at least add TODOs for that.

tarcieri:
This now implements parallel AES-CTR as well as interleaving AES-NI and CLMUL during decryption. I'm getting ~5cpb encryption and ~8cpb per decryption, which while not ideal are still in the right ballpark performance-wise.

Going to merge this soon unless there are any objections.

PR #12, opened 2019-09-27
add criterion benchmark for aes-gcm-siv

james-darkfox:
@tarcieri


tarcieri:
Haha, need some work before I really care to benchmark it, but sure, why not.

PR #21, opened 2019-10-04
Expose "detached" in-place encryption/decryption APIs

tarcieri:
Exposes an `encrypt_in_place_detached()` and `decrypt_in_place_detached()` API on all AEAD implementations in this repository.

Until we figure out what the upstream API looks like, this commit at least makes an in-place mode an option for those who don't want to allocate a buffer for every encryption/decryption operation.

### DIFF
--- a/aes-gcm-siv/Cargo.toml
+++ b/aes-gcm-siv/Cargo.toml
@@ -20,6 +20,7 @@ aes = "0.3"
 ctr = "0.3"
 polyval = "0.2"
 subtle = { version = "2", default-features = false }
+zeroize = { version = "1.0.0-pre", default-features = false }
 
 [dev-dependencies]
 criterion = "0.3.0"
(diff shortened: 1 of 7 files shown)


tarcieri:
I will open a separate PR proposing to add these methods to the `Aead` and `AeadMut` traits. I don't think they need to be the only in-place API, or even the preferred one for end users, but in terms of AEAD implementations, I hope this PR makes clear that they all benefit as I have used this pattern in all of the implementations.

tarcieri:
PR to add `encrypt_in_place_detached()` and `decrypt_in_place_detached()` to the `Aead` and `AeadMut` traits open here: https://github.com/RustCrypto/traits/pull/58

tarcieri:
Going to go ahead and land this, but I'll hold off on another release of any of these crates until I hear back on https://github.com/RustCrypto/traits/pull/58 and https://github.com/RustCrypto/traits/pull/59

PR #24, opened 2019-10-05
AES-SIV and AES-PMAC-SIV

tarcieri:
This is largely an import of the AES-(PMAC-)SIV implementation from Miscreant, taken from this commit:

https://github.com/miscreant/miscreant.rs/commit/71aa57519347d0b6a38204f86f2f0335756ae0b8

It uses the `AeadMut` API for now (but would be nice to convert to `Aead` eventually)

Note that the RFC 5297 description of the AEAD interface for AES-SIV cannot presently be expressed in terms of the `Aead`/`AeadMut` traits, since it supports a variable-length nonce (which is what Miscreant's current `Aead` trait also provides):

https://tools.ietf.org/html/rfc5297#section-6.1

> N_MIN  is 1 octet.
> N_MAX  is unlimited.

I have selected a 128-bit `NonceSize` based on the following SHOULD in RFC 5297:

https://tools.ietf.org/html/rfc5297#section-3

> If the nonce is random, it SHOULD be at least 128 bits in length

This is not ideal though, and means I had to delete the test vectors which use a nonce whose length is not 128-bits.

### DIFF
--- a/aes-siv/tests/vectors/aead.rs
+++ b/aes-siv/tests/vectors/aead.rs
@@ -1,0 +1,86 @@
+#![allow(dead_code)]
+
+pub use serde_json::Value as JsonValue;
+use subtle_encoding::hex;
+
+/// AES-SIV AEAD test vectors
+// TODO(tarcieri): extract these from JSON format
+const TEST_VECTORS: &str = r#"
+{
+    "examples:A<O>":[
+        {
+            "name:s":"AES-SIV Authenticted Encryption with Associated Data Example",
+            "alg:s":"AES-SIV",
+            "key:d16":"7f7e7d7c7b7a79787776757473727170404142434445464748494a4b4c4d4e4f",
+            "ad:d16":"00112233445566778899aabbccddeeffdeaddadadeaddadaffeeddccbbaa99887766554433221100",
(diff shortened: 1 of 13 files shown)


aes-siv/tests/vectors/aead.rs
```diff
@@ -0,0 +1,86 @@
+#![allow(dead_code)]
+
+pub use serde_json::Value as JsonValue;
+use subtle_encoding::hex;
+
+/// AES-SIV AEAD test vectors
+// TODO(tarcieri): extract these from JSON format
```
tarcieri: I would like to extract these vectors from the JSON before merging (using e.g. `hex-literal` instead)

aes-siv/tests/vectors/siv.rs
```diff
@@ -0,0 +1,246 @@
+pub use serde_json::Value as JsonValue;
+use subtle_encoding::hex;
+
+/// AES-CMAC-SIV test vectors
+// TODO(tarcieri): extract these from JSON format
```
tarcieri: Ditto for these

tarcieri:
Going to merge this and submit a follow-up PR to simplify the tests

PR #26, opened 2019-10-06
AES-GCM

tarcieri:
~~Presently failing some of the NIST CAVS vectors for AES-256-GCM (commented out and marked as TODO).~~ (fixed) All of the AES-128-GCM ones I've tried are working, though.

It'd probably be good to extract *all* of the CAVS vectors for a 16-byte tag length and include them in the tests.


tarcieri:
Looks like I broke `mulX_POLYVAL` with this change:

https://github.com/RustCrypto/universal-hashes/pull/28/files#diff-6189716e3235b5dfb266bd1c39947b1aR95

I locally reverted it and all of the test vectors are now passing.

tarcieri:
With the bug in the `ghash` crate fixed, it's passing all CAVS test vectors I've tried.

I'd like to get all of them (well, with a 16-byte tag anyway) into the test vectors before merging.

tarcieri:
Just pushed the full set of NIST CAVS vectors (for a 96-bit nonce and 128-bit tag) and they're passing locally.

tarcieri:
All NIST CAVS vectors are passing, so I'll call this good for a first pass.

PR #43, opened 2019-11-18
Upgrade aead to v0.2

devashishdxt:
Note: I was not able to upgrade aead for `aes-siv` crate. I'll probably need some help doing it. 

Also, in `xsalsa20poly1305`, I suspect `XSalsa20Poly1305::encrypt()` and `XSalsa20Poly1305::encrypt_in_place()` will give different results because, in `XSalsa20Poly1305::encrypt()`, `tag` is inserted before the encrypted data and `XSalsa20Poly1305::encrypt_in_place()` which is auto-implemented from `Aead` trait appends `tag` at the end of the buffer.

### DIFF
--- a/aes-gcm-siv/Cargo.toml
+++ b/aes-gcm-siv/Cargo.toml
@@ -19,7 +19,7 @@ categories = ["cryptography", "no-std"]
 maintenance = { status = "experimental" }
 
 [dependencies]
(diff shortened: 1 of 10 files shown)


tarcieri:
@devashishdxt awesome, thanks! This looks like a great start, and I can finish up the rest.

PR #51, opened 2019-11-27
Add a `heapless` feature to all crates + docs to use it

tarcieri:
Makes `heapless` a first-class feature on all crates, to simplify heapless usage (e.g. microcontrollers).

With `heapless` as a first-class feature, it becomes possible to explicitly document heapless usage as well, which this commit also does.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ rust:
   - nightly
 
 script:
-  - cargo test --all --release
+  - cargo test --all --release --lib
   - cargo test --all --all-features --release
 
 env:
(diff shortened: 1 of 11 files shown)


.travis.yml
```diff
@@ -6,7 +6,7 @@ rust:
   - nightly
 
 script:
-  - cargo test --all --release
+  - cargo test --all --release --lib
   - cargo test --all --all-features --release
```
tarcieri: This still tests the docs, which now require the `heapless` feature to build

PR #125, opened 2020-03-11
Handles variable-length nonce

### LINKED ISSUE #62: GCM: Allow nonces of any length
Currently, AesGcm only allows nonces with a length of 12 bytes. However, the [the AES-GCM specification](https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-38d.pdf) allows nonces of any length:

screenshot20191206010437

Is there a way to use `AesGcm` with a nonce of a length ≠ 12 bytes?

masihyeganeh:
I'm trying to fix #62, but I'm not sure I'm on the right track.
I read source code of other implementations of variable-length nonce GCM. They do it like this:

- if nonce is 12 bytes, copy the nonce in counter and add 1 to the last bit
- else copy GHASH of nonce to counter

I did that here. Maybe it's wrong. I'm not sure.
This will be a breaking change but as mentioned in [here](https://github.com/RustCrypto/traits/issues/43), it is time to introduce breaking changes.

### DIFF
--- a/aes-gcm/src/lib.rs
+++ b/aes-gcm/src/lib.rs
@@ -183,13 +183,12 @@ where
     B: BlockCipher<BlockSize = U16>,
     B::ParBlocks: ArrayLength<GenericArray<u8, B::BlockSize>>,
 {
-    type NonceSize = U12;
     type TagSize = U16;
     type CiphertextOverhead = U0;
 
     fn encrypt_in_place_detached(
         &self,
-        nonce: &GenericArray<u8, Self::NonceSize>,
+        nonce: &[u8],
         associated_data: &[u8],
         buffer: &mut [u8],
     ) -> Result<Tag, Error> {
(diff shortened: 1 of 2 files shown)


aes-gcm/src/lib.rs
```diff
@@ -183,23 +183,32 @@ where
...
     B::ParBlocks: ArrayLength<GenericArray<u8, B::BlockSize>>,
 {
-    type NonceSize = U12;
     type TagSize = U16;
     type CiphertextOverhead = U0;
 
     fn encrypt_in_place_detached(
         &self,
-        nonce: &GenericArray<u8, Self::NonceSize>,
+        nonce: &[u8],
```
tarcieri: This would require a change in the `Aead` trait. See https://github.com/RustCrypto/traits/issues/65

masihyeganeh: Yeah. I [did that](https://github.com/RustCrypto/traits/pull/85) too

tarcieri: Aah, apologies!

tarcieri:
@masihyeganeh so there's another option which preserves type safety for all nonce sizes and avoids making changes to the underlying trait... that would be making the AES implementation generic around different nonce sizes much in the same way it's already generic around `B::ParBlocks`.

I think that's probably the best way to go in general, actually. If you'd like I can adapt your PR to do do that, as `generic-array` and `typenum` are a bit mindbending.

masihyeganeh:
@tarcieri That would be great. let me know if I can help with that

tarcieri:
I've open a PR based on this one which makes `AesGcm` generic around the nonce size: #126

XinGOfCloude18 (approved):

PR #126, opened 2020-03-14
aes-gcm: make AesGcm generic around nonce size

tarcieri:
Support for non-96-bit nonces, while still implementing the current `Aead` interface and type safety for nonce lengths.

### DIFF
--- a/aes-gcm/src/ctr.rs
+++ b/aes-gcm/src/ctr.rs
@@ -27,6 +27,9 @@ where
 
     /// Current CTR value
     counter_block: Block128,
+
+    /// Base value of the counter
+    base_counter: u32,
 }
 
 impl<B> Ctr32<B>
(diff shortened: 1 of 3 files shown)


tarcieri:
The test vectors still pass for 96-bit nonces, however there aren't any for non-96-bit nonces, and I'm not convinced this is correct yet, namely this part:

Screen Shot 2020-03-14 at 9 32 15 AM

It appears after inputting a zero-padded non-96-bit nonce into GHASH, it should also be followed by its length.

tarcieri:
I added some tests for non-96-bit IVs based on the NIST CAVS vectors and they're presently failing, even after attempting to include the IV length in the GHASH message.

tarcieri:
Ok, got it working after updating the CTR implementation to parse a base counter value from 𝐉₀

PR #156, opened 2020-06-05
Move ring-aead ro RustCrypto/ring-wrappers

### LINKED ISSUE #154: Move ring-aead into a separate repository?
RustCrypto is generally for pure-Rust implementations, so `ring-aead` looks a bit out of place. We already have `asm-hashes`, so we are not completely puristic in this regard. Thus how about moving the crate into a separate repository? In future, if someone will be interested, we may add wrapper crates for other algorithm families as well.

newpavlov:
Closes #154


tarcieri:
Weird, not sure why the audit is failing...

newpavlov:
Yeah, it probably will get fixed when `pre` dependencies will be cleaned.

PR #174, opened 2020-06-12
Add CCM

newpavlov:
For now implementation does not use parallel block processing for CTR part. Also not all combinations of tag and nonce size are tested, only vectors from RFC 3610.

cc @martindisch


tarcieri:
Perhaps we could ask @martindisch if he'd be interested in upstreaming his existing `aes-ccm` crate?

martindisch:
Seeing as you've already put a lot of work into my crate and I don't always have the capacity to maintain it as proactively as I'd like to, that's something I'd be open to, provided there's a consensus on your part that this would be beneficial for the RustCrypto project (instead of a from-scratch implementation) and we can agree on the specifics of the handover. But you should consider that mine is a port of Intel's implementation from [TinyCrypt](https://github.com/intel/tinycrypt) at its core. I don't foresee any issues, as it's licensed quite permissively, but something to keep in mind nonetheless.

Going off-topic here just for a small update for you: I've found some AES256 test vectors and am going to incorporate them before the 0.5 release this weekend.

tarcieri:
@martindisch glad to hear it re: AES-256-CCM test vectors.

@newpavlov what do you think?

newpavlov:
This implementation is not only generic over block ciphers, but also allows to select both tag size and nonce size/max payload size. So IIUC `aes-gcm` will become essentially a convenience wrapper around `gcm` and `aes`, correct?

Since it's a clear room implementation based on the RFC, I would be interested in hearing how you think the code can be improved (right now I am not happy about several places), maybe we should incorporate some approaches used in `aes-gcm`?

@martindisch 
>I've found some AES256 test vectors and am going to incorporate them before the 0.5 release this weekend.

Can you link them? I would like to add them them to this PR as well.

ccm/src/lib.rs
```diff
@@ -0,0 +1,303 @@
...
+            b[0] = 0xFF;
+            b[1] = 0xEE;
+            b[2..6].copy_from_slice(&la_arr[la_arr.len() - 4..]);
+            Some((b, 6))
+        } else {
+            let mut b = Block::<C>::default();
+            b[0] = 0xFF;
+            b[1] = 0xFF;
+            b[2..10].copy_from_slice(&la_arr[la_arr.len() - 8..]);
+            Some((b, 10))
```
newpavlov, line 139: This branch is not covered by tests, but to fix that, we would have to test AD bigger than 4 GiB. With `aesni` as a backend in release mode it takes several seconds, but with `aes-soft` it takes more than a minute.

martindisch:
https://csrc.nist.gov/Projects/cryptographic-algorithm-validation-program/CAVP-TESTING-BLOCK-CIPHER-MODES
Look for the "CCM Test Vectors" download. In there you can find some in `DVPT256.rsp` and other files.

tarcieri:
@newpavlov will try to review it in more detail

tarcieri:
@newpavlov looks like it needs rustfmt

newpavlov:
Phew... I've added all CVAP vectors. Even with the new `blobby` version, vectors take 364 KB. I will merge this PR after publishing the new versions of `blobby` and `aead`.

ccm/src/traits.rs
```diff
@@ -0,0 +1,60 @@
...
+    fn get_l() -> u8 {
+        15 - Self::to_u8()
+    }
+
+    fn get_max_len() -> usize {
+        // a somewhat ugly code to prevent overlfow.
+        // compiler should be able to completely optimize it out
+        let l = Self::get_l() as u128;
+        let v = (1 << (8 * l)) - 1;
+        core::cmp::min(v, core::usize::MAX as u128) as usize
```
newpavlov, line 42: This part is a bit ugly, but I couldn't find a better option.

newpavlov:
@tarcieri 
I think this PR is ready for review. It would nice to get 100% coverage, but for now it is blocked on RustCrypto/block-ciphers#135.

PR #179, opened 2020-07-11
Add eax crate

Flakebi:
This adds the eax crate to this repository.
I wrote it a while back here: https://github.com/Flakebi/eax-rs

Do you think it makes sense to have it in this repository?


eax/src/lib.rs
```diff
@@ -0,0 +1,299 @@
...
+//! By default this crate will use software implementations of both AES and
+//! the POLYVAL universal hash function.
+//!
+//! When targeting modern x86/x86_64 CPUs, use the following `RUSTFLAGS` to
+//! take advantage of high performance AES-NI and CLMUL CPU intrinsics:
+//!
+//! ```text
+//! RUSTFLAGS="-Ctarget-cpu=sandybridge -Ctarget-feature=+aes,+sse2,+sse4.1,+ssse3"
+//! ```
+//!
```
tarcieri, line 15: Looks like some copypasta. Can you remove it? Thanks!

eax/src/lib.rs
```diff
@@ -0,0 +1,299 @@
...
+
+/// EAX with a 128-bit key and 96-bit nonce
+#[cfg(feature = "aes")]
+#[cfg_attr(docsrs, doc(cfg(feature = "aes")))]
+pub type EaxAes128 = Eax<Aes128>;
+
+/// EAX with a 256-bit key and 96-bit nonce
+#[cfg(feature = "aes")]
+#[cfg_attr(docsrs, doc(cfg(feature = "aes")))]
+pub type EaxAes256 = Eax<Aes256>;
```
tarcieri, line 134: I think typically the mode comes second after the cipher and these should be `AesEax128` and `AesEax256`.

tarcieri:
Thanks, this looks great!

Would you mind adding the test vectors from Appendix G of the paper? (p.43):

https://web.cs.ucdavis.edu/~rogaway/papers/eax.pdf

Flakebi:
Thanks for the link to the test vectors, I wondered if there are any but didn’t find much. I added them in the same way as they are in the gcm crate.

tarcieri:
Awesome, thanks!

Can you add me as a crate owner?

eax/src/lib.rs
```diff
@@ -0,0 +1,286 @@
...
+            .encrypt_in_place_detached(&[0; 16].into(), &[1, 2, 3, 4], &mut data)
+            .unwrap();
+        assert_eq!(
+            mac.as_slice(),
+            &[232, 88, 147, 206, 130, 126, 14, 121, 62, 127, 33, 233, 239, 81, 51, 177]
+        );
+        assert_eq!(&data, &[182, 81, 68, 170, 62]);
+    }
+
+    quickcheck! {
```
newpavlov, line 277: Do we want to property-based testing here? I think it will unnecessary make CI runs longer.

eax/tests/aes128eax.rs
```diff
@@ -0,0 +1,91 @@
...
+
+use self::common::TestVector;
+use eax::aead::{generic_array::GenericArray, Aead, NewAead, Payload};
+use eax::Aes128Eax;
+
+/// EAX test vectors
+///
+/// <https://web.cs.ucdavis.edu/~rogaway/papers/eax.pdf>
+const TEST_VECTORS: &[TestVector<[u8; 16]>] = &[
+    TestVector {
```
newpavlov, line 17: I think it would be better to use the `aead::dev::new_test` macro, but I was a bit late. I will create a follow-up PR.

eax/src/lib.rs
```diff
@@ -0,0 +1,286 @@
...
+        // (^ means xor)
+        let expected_tag = n.zip(h, |a, b| a ^ b).zip(c, |a, b| a ^ b);
+
+        let expected_tag = &expected_tag[..tag.len()];
+
+        // Check mac using secure comparison
+        use subtle::ConstantTimeEq;
+        if expected_tag.ct_eq(tag).unwrap_u8() == 1 {
+            // Decrypt
+            let mut cipher = ctr::Ctr128::<Cipher>::from_block_cipher(Cipher::new(&self.key), &n);
```
newpavlov, line 224: Note that this approach will result in a lower performance for AES compared to types from `aesni`, but I guess there is no good way to solve it right now...

Also does the paper specify counter width? Because implementation in the `ctr` crate uses 64-bit counters.

eax/src/lib.rs
```diff
@@ -0,0 +1,286 @@
...
+/// Maximum length of ciphertext
+pub const C_MAX: u64 = (1 << 36) + 16;
+
+/// EAX tags
+pub type Tag = GenericArray<u8, U16>;
+
+/// EAX with a 128-bit key and 96-bit nonce
+#[cfg(feature = "aes")]
+#[cfg_attr(docsrs, doc(cfg(feature = "aes")))]
+pub type Aes128Eax = Eax<Aes128>;
```
newpavlov, line 117: Adding such alias will be inconsistent with other crates. Should we replace it with `aes-eax` wrapper crate?

eax/LICENSE-MIT
```diff
@@ -0,0 +1,25 @@
+Copyright (c) 2019 The RustCrypto Project Developers
```
newpavlov, line 1: 2020

newpavlov:
I was a bit late. :)

tarcieri:
Aah whoops, sorry about that

PR #180, opened 2020-07-11
Fix benchmarks

Flakebi:
Building the benchmarks was broken. This fixes it.


tarcieri:
Thanks! I was aware of this, but unfortunately never filed an issue or fixed it.

It'd be good to `cargo build --benches` as part of the CI process to catch this sort of thing in advance, but in the meantime, thanks for fixing this!

PR #182, opened 2020-07-12
Minor EAX changes

newpavlov:
This PR contains the following changes:
- `aead::dev::new_test` macro is used for tests (i.e. test vectors stored in the blobby format)
- Removed `quickcheck` tests
- Removed AES type aliases for consistency with other crates (if we decide that such features are preferred to the wrapper crates, we could return them later)


newpavlov:
IIUC the paper does no specify counter width for the CTR mode. Ideally we should check what other implementations do.

PR #185, opened 2020-07-31
Add Multilinear Galois Mode

newpavlov:
For now it uses standalone code for simplification reasons, but in future hopefully it will use the same code base as `ghash` and `polyval`. I haven't ported `soft_u32` and probably I will postpone it until unification with `polyval`.

MGM can also work with 64-bit wide block ciphers, but I doubt I will bother implementing it.


mgm/src/gf/u64_soft.rs
```diff
@@ -0,0 +1,76 @@
...
+        block
+    }
+}
+
+/// Multiplication in GF(2)[X], truncated to the low 64-bits, with “holes”
+/// (sequences of zeroes) to avoid carry spilling.
+///
+/// When carries do occur, they wind up in a "hole" and are subsequently masked
+/// out of the result.
+fn bmul64(x: u64, y: u64) -> u128 {
```
newpavlov, line 47: Note that I am using `u128` here instead of the reversing trick. It allows to simplify code a bit. Maybe it's worth to port this approacht to `polyval`?

tarcieri: Sure.

(I'm also curious why all of this can't be implemented in terms of POLYVAL?)

newpavlov: >I'm also curious why all of this can't be implemented in terms of POLYVAL?

It could be, but I am not sure if it will have the same performance as an "inlined" version. Plus there are some annoying differences between algorithms, e.g. MGM does not chain values like GHASH and Polyval and it uses a "direct" order of bits for mapping binary string to polynomial. I will look into unifying code bases later, but not now.

tarcieri: Aah, ok!

mgm/src/lib.rs
```diff
@@ -0,0 +1,271 @@
...
+//! # Example
+//! ```
+//! # #[cfg(feature = "alloc")]
+//! # {
+//! use mgm::Mgm;
+//! use kuznyechik::Kuznyechik;
+//! use mgm::aead::{Aead, NewAead, generic_array::GenericArray};
+//!
+//! let key = GenericArray::from_slice(b"an example very very secret key.");
+//! let cipher = Mgm::<Kuznyechik>::new(key);;
```
kpp, line 12: double `;;` at the end of the list

tarcieri: Surprised `clippy` didn't catch that

newpavlov: Currently `clippy` does not check code examples, see: rust-lang/rust#56232

ethindp:
Does anyone know how secure this is? I can't seem to find any information on that, and I'm not good at reading papers on cryptographic algorithms or hash functions.

newpavlov:
AFAIK the construction was not properly analyzed by independent cryptographers yet. The authors claim some improvements compared to GCM, but in my understanding they are mostly theoretical. On a con side those properties are achieved by significantly degrading performance, if `m` is number of blocks in authenticated data  and `n` in plaintext, GCM ecryption requires `n+1` applications of a block cipher, while MGM requires `2*n + m + 3`, i.e. asymptotic performance is similar to [SIV](https://github.com/miscreant/meta/wiki/AES-SIV).

Contrary to MGM and GCM, SIV can not be fully parallelized (but it can be fixed by using [PMAC-SIV](https://github.com/miscreant/meta/wiki/AES-PMAC-SIV) construction) and requires two passes instead of one (assuming you are not saturating your memory bus it should not be a big problem thanks to [prefetcher](https://en.wikipedia.org/wiki/Cache_prefetching)), but it has a really nice property of [nonce reuse misuse resistance](https://github.com/miscreant/meta/wiki/Nonce-Reuse-Misuse-Resistance), which is far more important (as far as I understand) in practice than the theoretical improvements of MGM.

Also MGM authors made, let's say, an "interesting" decision to use nonces with length not multiple of 8 bits (127 and 63 bits for 128 and 64 bit block ciphers respectively), which adds some additional headache when generating nonces (current implementation will return an error if the most significant bit in 128 bit nonce is not equal to zero).

So if you don't target Russian crypto standards, I think it is better to choose between GCM and SIV modes.

tarcieri:
Also note: AES-GCM-SIV provides both parallelism and AES-GCM equivalent performance (with a ~20% encryption performance penalty, but decryption at equivalent-or-faster speeds to AES-GCM)

ethindp:
Yeah... the fact that it hasn't been independently reviewed makes it
sketchy. And the use of a non-power-of-two nonce is... really odd. I
was thinking of using GCM-SIV over standard GCM just for the nonse
reuse protection.

On 9/24/20, Tony Arcieri <notifications@github.com> wrote:
> Also note: AES-GCM-SIV provides both parallelism and AES-GCM equivalent
> performance (with a ~20% encryption performance penalty, but decryption at
> equivalent-or-faster speeds to AES-GCM)
>
> --
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub:
> https://github.com/RustCrypto/AEADs/pull/185#issuecomment-698675859


-- 
Signed,
Ethin D. Probst

Issue #184, opened 2020-07-31
A constant-time version of aes-gcm decrypt-in-place-detached?


garbageslam:
Hi, I have a use-case that is roughly the following (simplified):
- A private key exists only in an SGX enclave
- A stream of ciphertexts are passed to the enclave, which it attempts to decrypt
- It must remain a secret which ones were decryptable by the enclave and which ones aren't, taking into account side-channel leakage via code and data access patterns.

Unfortunately I can't quite do that with the aes-gcm code as is:

```
    fn decrypt_in_place_detached(
        &self,
        nonce: &GenericArray<u8, NonceSize>,
        associated_data: &[u8],
        buffer: &mut [u8],
        tag: &Tag,
    ) -> Result<(), Error> {
        if buffer.len() as u64 > C_MAX || associated_data.len() as u64 > A_MAX {
            return Err(Error);
        }

        // TODO(tarcieri): interleave encryption with GHASH
        // See: <https://github.com/RustCrypto/AEADs/issues/74>
        let mut expected_tag = self.compute_tag(associated_data, buffer);
        let mut ctr = self.init_ctr(nonce);
        ctr.apply_keystream(&self.cipher, expected_tag.as_mut_slice());

        use subtle::ConstantTimeEq;
        if expected_tag.ct_eq(&tag).unwrap_u8() == 1 {
            ctr.apply_keystream(&self.cipher, buffer);
            Ok(())
        } else {
            Err(Error)
        }
    }
```

What I would really like is a version of this function that looks like this for example:

```
    fn ct_decrypt_in_place_detached(
        &self,
        nonce: &GenericArray<u8, NonceSize>,
        associated_data: &[u8],
        buffer: &mut [u8],
        tag: &Tag,
    ) -> bool {
        if buffer.len() as u64 > C_MAX || associated_data.len() as u64 > A_MAX {
            return false;
        }

        // TODO(tarcieri): interleave encryption with GHASH
        // See: <https://github.com/RustCrypto/AEADs/issues/74>
        let mut expected_tag = self.compute_tag(associated_data, buffer);
        let mut ctr = self.init_ctr(nonce);
        ctr.apply_keystream(&self.cipher, expected_tag.as_mut_slice());
        ctr.apply_keystream(&self.cipher, buffer);

        use subtle::ConstantTimeEq;
        bool::from(expected_tag.ct_eq(&tag))
    }
```

The key differences being:
[...]

PR #188, opened 2020-08-05
Expose AesGcm bits necessary to build a constant time decrypt

xoloki:
As discussed in https://github.com/RustCrypto/AEADs/issues/184, aes-gcm decrypt is not constant time.  @tarcieri suggested that for our use case, it would be better to build a custom crate rather than put a constant time decrypt in the base repo, since overwriting the buffer if the mac check fails is necessary to avoid leaking information.

We attempted to build a constant time decrypt module without duplicating the entire aes-gcm crate; unfortunately, the AesGcm struct does not allow access to the critical data and functions.  So I'm offering this PR, which simply makes one struct field and two functions pub.

### DIFF
--- a/aes-gcm/src/ctr.rs
+++ b/aes-gcm/src/ctr.rs
@@ -14,7 +14,7 @@ type Block128 = GenericArray<u8, U16>;
 pub(crate) const BLOCK_SIZE: usize = 16;
 
 /// CTR mode with a 32-bit big endian counter
-pub(crate) struct Ctr32<Aes>
+pub struct Ctr32<Aes>
 where
     Aes: BlockCipher<BlockSize = U16>,
     Aes::ParBlocks: ArrayLength<GenericArray<u8, Aes::BlockSize>>,
--- a/aes-gcm/src/lib.rs
+++ b/aes-gcm/src/lib.rs
@@ -170,7 +170,7 @@ where
     NonceSize: ArrayLength<u8>,
 {
     /// Encryption cipher
-    cipher: Aes,
+    pub cipher: Aes,
 
     /// GHASH authenticator
     ghash: GHash,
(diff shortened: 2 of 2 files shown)


tarcieri:
I'd be okay with moving `Ctr32` somewhere it can be `pub`, possibly the `ctr` crate (it could be unified with the `Ctr32` used by AES-GCM-SIV, which is effectively the same thing, but little endian).

I'm somewhat reluctant to expose methods like `init_ctr` and `compute_tag` though, as these are effectively internals of the implementation.

tarcieri:
See this comment re: extracting `Ctr32` into the `ctr` crate:

https://github.com/RustCrypto/stream-ciphers/issues/41

If you had that, would you be okay with duplicating the rest of the crate? Especially if you went back to a slightly earlier version of `aes-gcm` before support for variable-sized nonces was added, the bulk of the remaining implementation is ~80 LOC, much of which I imagine you'd want to customize to get the behavior you desire:

https://github.com/RustCrypto/AEADs/blob/5a28e702bc1019115154521273710568889b89f6/aes-gcm/src/lib.rs#L181-L254

xoloki:
> If you had that, would you be okay with duplicating the rest of the crate? 

Yeah that's cool, thanks!

tarcieri:
PR to extract `Ctr32` into the `ctr` crate (as `Ctr32BE`):

https://github.com/RustCrypto/stream-ciphers/pull/170

If that's sufficient and you can do the rest yourselves in a 3rd party crate, I'll go ahead and close this.

tarcieri:
Note: in `ctr` v0.6.0-pre, there's now `Ctr32BE`: https://docs.rs/ctr/0.6.0-pre/ctr/struct.Ctr32BE.html

Issue #132, opened 2020-05-18
Stream data to cipher


acasajus:
Hello,

 Is it possible to stream in data to cipher to avoid having to have all the data in memory to pass it though the AEAD? Something like (in pseudocode)

```
 aead = AEAD::new(mode, nonce)
 while plaindata = plainfile.read(1024) {
  cipherfile.write(aead.encrypt(plaindata))
}
cipherfile.write(aead.close_and_flush())
```

How could I do this if the file is several gigs in size and I don't have that much mem?

Thanks in advance,

tarcieri:
Streaming *encryption* like this is possible with some AEAD modes, depending on how the algorithm is structured. It requires a single-pass construction that places things like the AAD and plaintext length at the end of the MAC input.

Note that streaming decryption which exposes candidate plaintexts without first checking the MAC tag is inherently unsafe.

The general security framework for this concept I prefer for solving this particular problem [Online Authenticated Encryption (OAE)](https://web.cs.ucdavis.edu/~rogaway/papers/oae.pdf) and within that framework there are two provably secure schemes for this developed by Phil Rogaway et al, one of which I've implemented in [Miscreant](https://miscreant.io/) which is the predecessor of the `aes-siv` crate:

- [STREAM](https://github.com/miscreant/meta/wiki/STREAM): provides nonce-based online authenticated encryption (nOAE), a segmented/streaming AEAD scheme
- [CHAIN](https://github.com/miscreant/meta/issues/33): (mentioned in the aforementioned paper) requires an MRAE cipher but provides a stronger notion of security called OAE2 which provably guarantees a sequence of AEAD messages was encrypted/decrypted in-order

Note that [`STREAM` is presently implemented in Miscreant](https://docs.rs/miscreant/0.4.2/miscreant/stream/index.html). It might be interesting to upstream its implementation somewhere in the RustCrypto project, which is something I've considered in the past but haven't actually done yet.

acasajus:
Thanks for the comprehensive response. I'll take a look at [Miscreant](https://miscreant.io/) :+1:

tarcieri:
As of `aead` v0.4 there is now an [`aead::stream`](https://docs.rs/aead/latest/aead/stream/index.html) module.

Issue #74, opened 2020-01-26
One-pass encryption/decryption


tarcieri:
Currently all of the AEAD implementations do two passes over the plaintext/ciphertext when encrypting/decrypting respectively: for encryption, they encrypt the plaintext in the first pass, and authenticate it in the second pass. For decryption, it's vice versa.

A better approach is to pick a number of blocks to operate on in parallel and encrypt/authenticate or authenticate/decrypt in a single pass. This has better cache locality, e.g. when we encrypt data, store the resulting ciphertext, then load it again to do authentication, that is pretty much guaranteed to hit L1 cache when doing it in a single pass (and ideally we could hand off values still stored in e.g. SIMD registers)

This is a tracking issue for converting the implementations of these respective algorithms to be one pass. It also might be good to discuss ways we could have a generic implementation of one pass encryption/decryption in the `aead` crate (especially one specialized for the non-SIV `stream-cipher` + `universal-hash` use case) which can be reused across different algorithm implementations.

- [ ] `aes-gcm`
- [ ] `aes-gcm-siv`
- [ ] `aes-siv`
- [ ] `chacha20poly1305`
- [ ] `xsalsa20poly1305`

tarcieri:
Related issue: generalized AEAD implementations based on stream ciphers https://github.com/RustCrypto/traits/issues/45

nico-abram:
As a complete crypto noob who got here from github explore (But *not* a complete rust noob), would this be feasible? (Or even just a small part, like trying to make decryption for chacha20poly1305 single pass) How robust are existing tests?

Would the change mostly be changing implementations like https://github.com/RustCrypto/AEADs/blob/master/chacha20poly1305/src/cipher.rs#L66-L91 into something more like https://github.com/RustCrypto/AEADs/blob/master/aes-gcm-siv/src/lib.rs#L317-L347 ?

tarcieri:
Yes, but it also needs to be done in a way that actually improves performance. I've tried to do this change naively a few times (to `aes-gcm`, mainly, I might still have the code around) and it decreased performance.

I think doing it properly might require keeping the data flowing through XMM registers... at the very least it needs to all stay in L1 cache.

nico-abram:
Thanks for the response!

I think I'll give it a shot. Do not let that stop anyone else from trying it since I don't have much hope I'll be able to do much.

[...]

PR #214, opened 2020-09-09
eax: Introduce a streaming API

Xanewok:
Introduces a ` stream` module that implements `EaxStream`, which is the streaming variant of the regular `Eax` implemented here. 

This can be useful for two reasons, in addition to having the existing one-shot-like API:
1. The associated data can be processed separately
2. The protected data can be processed in a streaming fashion, rather than requiring it to be entirely allocated and contiguous.

I'm trying to reuse the EAX implementation for use in [Sequoia](https://gitlab.com/sequoia-pgp/sequoia) (Rust OpenPGP implementation) but unfortunately I need the streaming API variant for that.

Because the newly-introduced in-place API can be somewhat easily misused, I tried to mimic a linear type, forcing the user to explicitly consume and handle the resulting value after decryption. With `std` feature enabled, it aborts the process using a drop bomb pattern, unless it's forgotten by the `EaxStream::finish` associated function.

Note that this changes the existing `Eax` implementation by first performing the decryption in-place and then checking against the expected tag, rather than scanning the ciphertext first to calculate tag and then decrypting if the tags match. I have to admit that I did that to share more of the implementation and leverage the testing framework test both implementations but I can revert that change if needed.

Let me know what you think about this and if this is something you'd consider merging in one form or another.

### DIFF
--- a/eax/src/online.rs
+++ b/eax/src/online.rs
@@ -37,9 +37,9 @@
 //! // Now decrypt it, using the same key and nonce
 //! let mut cipher = Eax::<Aes256, Decrypt>::with_key_and_nonce(key, nonce);
 //! cipher.update_assoc(&assoc[..]);
-//! cipher.decrypt_unauthenticated(&mut buffer[..5]);
-//! cipher.decrypt_unauthenticated(&mut buffer[5..10]);
-//! cipher.decrypt_unauthenticated(&mut buffer[10..]);
+//! cipher.decrypt_unauthenticated_hazmat(&mut buffer[..5]);
+//! cipher.decrypt_unauthenticated_hazmat(&mut buffer[5..10]);
+//! cipher.decrypt_unauthenticated_hazmat(&mut buffer[10..]);
 //! let res = cipher.finish(&tag);
 //!
 //! assert_eq!(res, Ok(()));
@@ -48,7 +48,7 @@
 //! // Decrypting the ciphertext with tampered associated data should fail
 //! let mut cipher = Eax::<Aes256, Decrypt>::with_key_and_nonce(key, nonce);
 //! cipher.update_assoc(b"tampered");
-//! cipher.decrypt_unauthenticated(&mut cloned);
+//! cipher.decrypt_unauthenticated_hazmat(&mut cloned);
 //! let res = cipher.finish(&tag);
 //!
 //! assert_eq!(res, Err(Error));
@@ -121,9 +121,9 @@ impl CipherOp for Decrypt {}
 /// // Now decrypt it, using the same key and nonce
 /// let mut cipher = Eax::<Aes256, Decrypt>::with_key_and_nonce(key, nonce);
 /// cipher.update_assoc(&assoc[..]);
-/// cipher.decrypt_unauthenticated(&mut buffer[..5]);
-/// cipher.decrypt_unauthenticated(&mut buffer[5..10]);
-/// cipher.decrypt_unauthenticated(&mut buffer[10..]);
+/// cipher.decrypt_unauthenticated_hazmat(&mut buffer[..5]);
+/// cipher.decrypt_unauthenticated_hazmat(&mut buffer[5..10]);
+/// cipher.decrypt_unauthenticated_hazmat(&mut buffer[10..]);
 /// let res = cipher.finish(&tag);
 ///
 /// assert_eq!(res, Ok(()));
@@ -133,7 +133,7 @@ impl CipherOp for Decrypt {}
 /// let mut cipher = Eax::<Aes256, Decrypt>::with_key_and_nonce(key, nonce);
 ///
 /// cipher.update_assoc(b"tampered");
-/// cipher.decrypt_unauthenticated(&mut cloned);
+/// cipher.decrypt_unauthenticated_hazmat(&mut cloned);
 /// let res = cipher.finish(&tag);
 ///
 /// assert_eq!(res, Err(Error));
@@ -221,9 +221,25 @@ where
     /// To correctly verify the authenticity, use the [`finish`] associated
     /// function.
     ///
+    /// # ☣️ BEWARE! ☣️
+    /// This is a low-level operation that simultaneously decrypts the data and
+    /// calculates an intermediate tag used to verify the authenticity of the
+    /// data (used when the online decryption is finished).
+    ///
+    /// Because this is exposed solely as a building block operation, an extra
+    /// care must be taken when using this function.
+    ///
+    /// Specifically, when misused this may be vulnerable to a chosen-ciphertext
+    /// attack (IND-CCA). Due to online nature of this function, the decryption
+    /// and partial tag calculation is done simultaneously, per chunk.
+    /// An attacker might choose ciphertexts to be decrypted and, while the
+    /// final decryption will fail because the attacker can't calculate tag
+    /// authenticating the message, obtained decryptions may leak information
+    /// about the decryption scheme (e.g. leaking parts of the secret key).
+    ///
     /// [`finish`]: #method.finish
     #[inline]
-    pub fn decrypt_unauthenticated(&mut self, msg: &mut [u8]) {
+    pub fn decrypt_unauthenticated_hazmat(&mut self, msg: &mut [u8]) {
         self.imp.decrypt(msg)
     }
 
--- a/eax/src/online.rs
+++ b/eax/src/online.rs
@@ -10,10 +10,6 @@
 //! This will either return a *tag* (when encrypting) used to authenticate data
 //! or a `Result` (when decrypting) that signifies whether the data is authentic,
 //! which is when the resulting tag is equal to the one created during encryption.
-//! # Panic
-//! If the `Eax` value will not be consumed via [`finish`] the
-//! process will abort, when compiled with `std` feature enabled, to prevent
-//! against bugs related to decrypting data without verifying its authenticity.
 //!
 //! ## Example
 //! ```
@@ -65,7 +61,6 @@
 use crate::*;
 
 use core::marker::PhantomData;
-use core::mem;
 
 pub use Eax as EaxOnline;
 
@@ -97,10 +92,6 @@ impl CipherOp for Decrypt {}
 /// or a `Result` (when decrypting) that signifies whether the data is authentic,
 /// which is when the resulting tag is equal to the one created during encryption.
 ///
-/// # Panic
-/// If the `Eax` value will not be consumed via [`finish`] the
-/// process will abort, when compiled with `std` feature enabled, to prevent
-/// against bugs related to decrypting data without verifying its authenticity.
 /// ## Example
 /// ```
 /// use eax::{Error, online::{Eax, Decrypt, Encrypt}};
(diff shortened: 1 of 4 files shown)


Xanewok:
codecov.io check fails with
```
Please provide an upload token from codecov.io with valid arguments
```

tarcieri:
Some previous discussion on #132.

One tricky naming thing here is we intend to implement [STREAM](https://github.com/miscreant/meta/wiki/STREAM), a provably secure segmented AEAD (a.k.a. OAE) construction designed by Phil Rogaway. This can be implemented generically for any AEAD mode (a.k.a. nonce-based OAE).

As a contrast to that, perhaps we could find another word for the concept in this PR (i.e. #132) within the RustCrypto/AEAD implementations. Perhaps "incremental encryption"?

tarcieri:
> codecov.io check fails

Will fix, sorry about that.

Xanewok:
Thanks for taking a look! I pushed a commit that prefers using the "online" word rather than stream.

I'm kind of torn about not implementing [`AeadMutInPlace`](https://docs.rs/aead/0.3.2/aead/trait.AeadMutInPlace.html) directly. One of the reasons is that we shouldn't need to pass the `nonce` to the encrypt/decrypt function everytime, once you instantiate the EAX mode. Another one is that the chunk decryption will always perform tag calculation but that may be wasteful and might be hard for the optimizer to recognize. It looks like this interface may not be designed for streaming/online use?

Do you think I should also implement  [`AeadMutInPlace`](https://docs.rs/aead/0.3.2/aead/trait.AeadMutInPlace.html) for the newly-introduced `EaxOnline`?

tarcieri:
> I pushed a commit that prefers using the "online" word rather than stream.

"Online" is also problematic, as STREAM (as in the Rogaway version) is a mode of [Online Authenticated Encryption](https://eprint.iacr.org/2015/189.pdf) (specifically a nonce-based OAE or "nOAE").

Xanewok:
As much as I understand reserving the "STREAM" name because it's introduced as a name/term for a useful generic framework capable of "online-ifying" a given nAE scheme, I don't think we should also reserve the "online" noun. After all, it's a well-known name of a property of an algorithm, just like EAX is online by design in general.

Hypothetically, if I were to find anything related to the OAE/CHAIN/STREAM framework, I'd expect to find it under "OAE{,1,2}" as a, for better or worse, yet another introduced term in the nomenclature of authenticated encryption.

Xanewok:
Maybe "chunk(ed)" name would be better, then? Hopefully that would not imply the data to be of equal-length chunks, since "block" is a more commonly used name for this concept here.

tarcieri:
Regardless of what name we choose, I think it might be worth designing a trait for this use case and adding it to https://github.com/RustCrypto/traits/tree/master/aead

/cc @newpavlov

eax/src/online.rs
```diff
@@ -0,0 +1,326 @@
...
+/// Runtime-enforced linear-ish type.
+///
+/// This type is useful to enforce that a value is correctly explicitly consumed.
+/// Otherwise, this will abort (only when `std` feature is enabled).
+struct DropBomb;
+#[cfg(feature = "std")]
+impl Drop for DropBomb {
+    fn drop(&mut self) {
+        std::eprintln!("Drop bomb says buh-bye");
+        std::process::abort();
```
newpavlov, line 181: While I understand the purpose of this code, I am not really comfortable with it. I would prefer to mark "online AEAD" as a hazmat operation. We already (intentionally) expose a fair amount of operations, which could be dangerously misused without a proper knowledge, so I don't think we should add such fragile guard rails.

newpavlov (commented):
I am also not sure about a good name for such mode of operation. Maybe we could employ the IUF terminology here?

I think ideally we should design a trait for "online mode" first and update the crates later. There are some open design questions regarding buffering, enforcing encryptor/decryptor usage via type system and parallel processing of packets, but overall changes should be quite straightforward.

@Xanewok 
Would you be willing to wait until such changes and use your fork in meantime? If you plan to use it in Sequoia in the near future, I guess we could merge this variant for now and push a minor release and migrate to the new traits in the next major release.

eax/src/online.rs
Xanewok: I don't mind and would be fine with exposing the online set of operations as unsafe (in the general sense). Originally, I thought that the harder-to-misuse API approach would be preferred, so I erred on the safe side and wrote this, instead.

I'll remove the drop bomb code.

Xanewok:
> Would you be willing to wait until such changes and use your fork in meantime? If you plan to use it in Sequoia in the near future, I guess we could merge this variant for now and push a minor release and migrate to the new traits in the next major release.

We currently use our own [implementation](https://gitlab.com/sequoia-pgp/sequoia/-/blob/033540d9708b0f1de61572bf43212368599d891a/openpgp/src/crypto/backend/cng/aead.rs#L35-107) but we decided that upstreaming the implementation would be a better idea and less frowned upon by the users AIUI. I believe that it'd be great for us to merge the approved version of this PR and use a newly-published minor release, as we'd like to ship our own 1.0 soon (cc @nwalfield @teythoon for release details)

> There are some open design questions regarding buffering, enforcing encryptor/decryptor usage via type system and parallel processing of packets (...)

@newpavlov Do you have any concrete ideas that I could apply already to the API surface introduced here?

newpavlov:
>Do you have any concrete ideas that I could apply already to the API surface introduced here?

You already get it for free by using the `ctr` crate and CMAC can not be parallelized by design. The implementation could benefit from interleaving MAC update and application of the keystream (one-pass approach is cache-friendlier, than the currently used two-pass one, see #74), but it will not be trivial, since we do not want `ctr` to touch its inner buffer. Probably we will need deeper changes in our API to make such optimization more ergonomic, e.g. by exposing primitives which would do updates on blocks only without any inner buffering, on top of which will be built buffering and "all or nothing" counterparts.

nwalfield:
> > Would you be willing to wait until such changes and use your fork in meantime? If you plan to use it in Sequoia in the near future, I guess we could merge this variant for now and push a minor release and migrate to the new traits in the next major release.

Using a git fork is difficult, because as long as we are using a git fork, we can't make a release to crates.io, which, AIUI, only allows crates that rely on other crates published on crates.io.

> We currently use our own [implementation](https://gitlab.com/sequoia-pgp/sequoia/-/blob/033540d9708b0f1de61572bf43212368599d891a/openpgp/src/crypto/backend/cng/aead.rs#L35-107) but we decided that upstreaming the implementation would be a better idea and less frowned upon by the users AIUI.

There are a couple of things going on here.  First, we don't want to use our own crypto, even if it is only a few dozen lines of code.  Second, if we rely on something else, it would be good that it not pull in too many dependencies.  That is, if we use the eax crate to avoid including 50 lines of crypto code and eax pulls in x000 LOC, it's unclear that we've made the right tradeoff.

> I believe that it'd be great for us to merge the approved version of this PR and use a newly-published minor release, as we'd like to ship our own 1.0 soon (cc @nwalfield @teythoon for release details)

We're getting very close to a 1.0 release.  But, if we don't make it in the next few weeks, we'll almost certainly do another point release.  So, we can't merge the eax code until the eax crate is released.

eax/src/online.rs
```diff
@@ -0,0 +1,326 @@
...
+where
+    Cipher: BlockCipher<BlockSize = U16> + NewBlockCipher + Clone,
+    Cipher::ParBlocks: ArrayLength<Block<Cipher>>,
+{
+    /// Applies decryption to the ciphertext.
+    #[inline]
+    pub fn decrypt(&mut self, msg: &mut [u8]) {
+        self.message.update(msg);
+        self.ctr.apply_keystream(msg);
+    }
```
tarcieri, line 308: Like @newpavlov mentioned above, at the very least this needs to be market as hazmat (e.g. `decrypt_unauthenticated_hazmat` with a comment explaining how potential misuse can lead to catastrophic failures like chosen ciphertext attacks and potentially full plaintext recovery)

tarcieri:
I'd be okay with merging this for now after `DropBomb` is removed and the unauthenticated decryption method is renamed to note it's `hazmat` and has an appropriately descriptive comment about how potential misuse could lead to catastrophic cryptographic failures.

Xanewok:
Dropped the drop bomb code.

Before we merge this, I wanted to bring back the old behavior of first checking the tag and only then decrypting on success, in the offline variant.

This would, however, separate the offline/online implementations but I don't feel comfortable not testing the online variant at all. Unfortunately, the `aead` test case works in the integration tests format but I would not like to expose the raw set of operations and instead only expose the online mode to be used in either decrypt or encrypt mode. But because these are integration tests, I can't test private API nor I can `doc(hidden)` because of the trait interface I need to keep around a `Key` instance solely for tests...

Do you have any idea or preference on how to tackle this for the time being?

Should I adapt the `aead::new_test!` macro to accept a path to test case blob, rather than defaulting to `data/$test_case.blb`? This would allow me to define a test using that macro but from within the library unit tests, without exposing the internals.

eax/src/online.rs
```diff
@@ -0,0 +1,432 @@
...
+{
+    /// Applies decryption to the ciphertext **without** verifying the
+    /// authenticity of decrypted message.
+    ///
+    /// To correctly verify the authenticity, use the [`finish`] associated
+    /// function.
+    ///
+    /// [`finish`]: #method.finish
+    #[inline]
+    pub fn decrypt_unauthenticated(&mut self, msg: &mut [u8]) {
```
tarcieri, line 226: Would still prefer `hazmat` somewhere in here. IMO it's not unclear this can break IND-CCA(3) security.

eax/src/online.rs
```diff
@@ -0,0 +1,432 @@
...
+        self.imp.decrypt(msg)
+    }
+
+    /// Finishes the decryption stream, verifying whether the associated and
+    /// decrypted data stream has not been tampered with.
+    ///
+    /// This *must* be called after the stream decryption is finished.
+    #[must_use = "decrypted data stream must be verified for authenticity"]
```
tarcieri, line 234: `Result` is already `#[must_use]` so I don't think this accomplishes anything?

I think it'd also be good to talk about the breakages that failure to call `finish` might cause, but those should also be mentioned in `decrypt_unauthenticated`.

eax/src/online.rs
```diff
@@ -0,0 +1,432 @@
...
+    Cipher: BlockCipher<BlockSize = U16> + NewBlockCipher + Clone,
+    Cipher::ParBlocks: ArrayLength<Block<Cipher>>,
+{
+    nonce: Nonce<Cipher::BlockSize>,
+    data: Cmac<Cipher>,
+    message: Cmac<Cipher>,
+    ctr: ctr::Ctr128<Cipher>,
+    // HACK: Needed for the test harness due to AEAD trait online/offline interface mismatch
+    #[cfg(test)]
+    key: Key<Cipher>,
```
tarcieri, line 256: Hmm, it'd really be good if we could figure out some other solution to this.

eax/src/online.rs
Xanewok: That's to add the specific note. Do you want me to remove it?

tarcieri: If that actually works, we can keep it. Does it?

tarcieri (approved):
Approved, albeit with at least one notable open question: https://github.com/RustCrypto/AEADs/pull/214/files#r494580906

eax/src/online.rs
Xanewok: Yep, simply calling
```rust
cipher.finish(&tag);
```
yields the following warning:
```
warning: unused return value of `eax::online::Eax::<Cipher, eax::online::Decrypt>::finish` that must be used
  --> eax/tests/lol.rs:30:5
   |
30 |     cipher.finish(&tag);
   |     ^^^^^^^^^^^^^^^^^^^^
   |
   = note: decrypted data stream must be verified for authenticity
```

Do you want me to keep it or reword it?

eax/src/online.rs
Xanewok: I did add 3b180b9c0927e599e8d396ff8683f6b1bfeb05c3, which hopefully addresses this. I'm not a cryptographer so I'm not really sure that I understood/phrased the IND-CCA security in this context; let me know if the current phrasing is correct and acceptable.

eax/src/online.rs
tarcieri: That's fine

eax/src/online.rs
tarcieri: Looks good, thanks!

eax/src/online.rs
tarcieri: I guess this is my one lingering concern but I don't consider it a showstopper

tarcieri:
@newpavlov WDYT?

newpavlov (approved):
Looks good as a temporary solution, I think we can finalize the naming after adding streaming API to the `aead` crate.

@tarcieri 
Can you do a release after merge?

tarcieri:
Released in v0.2.0

Xanewok:
Thanks!

PR #231, opened 2020-10-17
eax: Allow variable tag length

### LINKED ISSUE #220: Partial tag comparison
Hi,
I’m trying to use the `eax` crate to decrypt data with `decrypt_in_place_detached` but I only have the first 8 byte of the tag available as the network protocol I implement cuts off the other 8 byte of the 16 byte tag.
All aead functions expect the whole tag, so I can’t find a way to decrypt the data.

Flakebi:
Implemented analogous to ccm by passing a second generic argument to
Eax.

Fixes #220.

This change should be mostly backwards compatible, apart from exporting the generic Tag instead of `pub type Tag = GenericArray<u8, U16>;`
It should be easy to make keep it completely backwards compatible if that’s better.

### DIFF
--- a/eax/src/traits.rs
+++ b/eax/src/traits.rs
@@ -1,0 +1,18 @@
+use aead::consts::{U16, U4};
+use aead::generic_array::typenum::type_operators::{IsGreaterOrEqual, IsLessOrEqual};
+use aead::generic_array::typenum::Unsigned;
+use aead::generic_array::ArrayLength;
+
+mod private {
+    // Sealed traits stop other crates from implementing any traits that use it.
+    pub trait SealedTag {}
+
+    impl<T> SealedTag for T where
+        T: super::IsGreaterOrEqual<super::U4> + super::IsLessOrEqual<super::U16>
+    {
+    }
+}
+
+pub trait TagSize: ArrayLength<u8> + Unsigned + private::SealedTag {}
+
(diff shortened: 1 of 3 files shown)


tarcieri (approved):

eax/src/traits.rs
```diff
@@ -0,0 +1,43 @@
...
+    impl SealedTag for super::U14 {}
+    impl SealedTag for super::U15 {}
+    impl SealedTag for super::U16 {}
+}
+
+pub trait TagSize: Unsigned + private::SealedTag {}
+
+impl TagSize for U1 {}
+impl TagSize for U2 {}
+impl TagSize for U3 {}
```
newpavlov, line 30: Are we sure we want to allow such short tags? Also I think it should be possible to specify a range bound using `typenum` without introducing the sealed trait (see [`typenum::type_operators`](https://docs.rs/typenum/1.12.0/typenum/type_operators/index.html)), though the current approach is simple and easier to understand, so we can leave it as-is.

newpavlov (approved):
As a temporary solution it looks good to me, but in future I would prefer to have the "hazmat" methods discussed in the parent issue.

Flakebi:
I changed the TagSize to be `>= U4` and `<= U16` and also put back the original `eax::Tag` alias, so the change should be backwards compatible.

tarcieri:
@Flakebi LGTM, and matches the `ccm` crate

PR #252, opened 2020-12-31
Bump `aes` crate; `aes*` crates => MSRV 1.49+

tarcieri:
The `aes` crate is a key component of several crates in this repo, and just had an MSRV breaking change in RustCrypto/block-ciphers#216.

This commit bumps the MSRV of all crates in this repo which use the `aes` crate to MSRV 1.49+.

In the case of `aes-gcm` and `aes-gcm-siv` specifically, the changes in RustCrypto/universal-hashes#113 bumped the MSRV of `ghash` and `polyval` to 1.49+ as well, so there are multiple motivations for this new MSRV.

### DIFF
--- a/.github/workflows/aes-gcm-siv.yml
+++ b/.github/workflows/aes-gcm-siv.yml
@@ -22,7 +22,7 @@ jobs:
     strategy:
       matrix:
         rust:
-          - 1.41.0 # MSRV
+          - 1.49.0 # MSRV
           - stable
         target:
           # - armv7-none-eabi TODO: test after MSRV is bumped to 1.42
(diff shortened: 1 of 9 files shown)


tarcieri:
Note: I know bumping MSRV to a Rust version released today is aggressive, but I'm not in any hurry to release any of these crates.

I think it'd be good to spend time making sure `cipher` is in a good state before releasing a v0.3, and maybe look at doing another breaking release of `universal-hash` (and potentially `aead`) as well.

This repo is sort of an integration capstone for all of the crates, so I expect it will take awhile to sort all of the above out, by which point Rust 1.50 may be out and Rust 1.49 will seem less onerous.

PR #263, opened 2021-01-11
ccm: fix panic on 32-bit targets and add CI for them

### LINKED ISSUE #258: ccm panics on 32 bit architectures
In some places liks this, `to_be_bytes` is called on `usize` values, which makes the code machine dependent:
https://github.com/RustCrypto/AEADs/blob/master/ccm/src/lib.rs#L119
If `l_arr` is only 4 bytes long, the index `q` a few lines below may overflow depending on the nonce length, for example with `N=7` (`l=8`).

tarcieri:
Runs tests on 32-bit Linux

_upd(newpavlov):_
Closes #258


newpavlov:
@tarcieri 
I have pushed the panic fix as a commit into this PR. Can you review it? Unless you have any objections/suggestions, I think we are good to merge.

cc @rkreis

tarcieri:
Looks good

PR #304, opened 2021-05-04
WIP: Legacy ChaCha20Poly1305 Construction

kamulos:
I just quickly copied the code from the `xchacha20poly1305` module and adapted it to `ChaCha20Legacy` together with the `LegacyNonce`. I set the `NonceSize` to `U8`.

It is however not working and I could use some hints where to search. When I encrypt using libsodium and decrypt using my new module the mac verification fails (`if self.mac.verify(tag).is_ok()` in `cipher.rs:87`). When I do the same thing with XChaCha20Poly1305 everything works. So probably I just did not change something I am supposed to change, but I don't know what that is.

This is my test program:

Cargo.toml
``` toml
[package]
name = "chacha"
version = "0.1.0"
edition = "2018"

[dependencies]
sodiumoxide = "0.2"
chacha20poly1305 = { path = "../AEADs/chacha20poly1305", features = ["legacy"] }
```

main.rs (comment in and out to switch to xchacha)
``` rust
// use sodiumoxide::crypto::aead::xchacha20poly1305_ietf as sodium_algo;
// const NONCE: &[u8; 24] = b"123456781234567812345678";
// type RustCryptoChaCha = chacha20poly1305::XChaCha20Poly1305;
// type RustCryptoNonce = chacha20poly1305::XNonce;

use sodiumoxide::crypto::aead::chacha20poly1305 as sodium_algo;
const NONCE: &[u8; 8] = b"12345678";
type RustCryptoChaCha = chacha20poly1305::ChaCha20Poly1305Legacy;
type RustCryptoNonce = chacha20poly1305::LegacyNonce;

use chacha20poly1305::aead::{Aead, NewAead};

fn main() {
    sodiumoxide::init().unwrap();
    let k = sodium_algo::gen_key();
    let n = sodium_algo::Nonce::from_slice(NONCE).unwrap();

    let cipher = sodium_algo::seal(b"Hello World!", None, &n, &k);

    println!(
        "sodium: {}",
        std::str::from_utf8(&sodium_algo::open(&cipher, None, &n, &k).unwrap()).unwrap()
    );

    let decrypter = RustCryptoChaCha::new_from_slice(&k.0).unwrap();
    let plaintext = decrypter
        .decrypt(RustCryptoNonce::from_slice(&n.0), &cipher[..])
        .unwrap();

    println!(
        "rust crypto: {}",
        std::str::from_utf8(&plaintext[..]).unwrap()
    );
}
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -142,6 +142,7 @@ dependencies = [
  "chacha20",
  "cipher",
  "poly1305",
+ "subtle",
  "zeroize",
 ]
 
--- a/chacha20poly1305/Cargo.toml
+++ b/chacha20poly1305/Cargo.toml
@@ -23,6 +23,7 @@ chacha20 = { version = "0.7", features = ["zeroize"], optional = true }
 cipher = "0.3"
 poly1305 = "0.7"
 zeroize = { version = "1", default-features = false }
+subtle = { version = "2", default-features = false }
 
 [dev-dependencies]
 aead = { version = "0.4", features = ["dev"], default-features = false }
@@ -35,6 +36,7 @@ heapless = ["aead/heapless"]
 reduced-round = ["chacha20"]
 stream = ["aead/stream"]
 xchacha20poly1305 = ["chacha20/xchacha"]
+legacy = ["chacha20/legacy"]
 
 [package.metadata.docs.rs]
 all-features = true
--- a/chacha20poly1305/src/chacha20poly1305legacy.rs
+++ b/chacha20poly1305/src/chacha20poly1305legacy.rs
@@ -1,0 +1,206 @@
+pub use chacha20::LegacyNonce;
+
+use crate::{Key, Tag};
+use aead::{
+    consts::{U0, U16, U32, U8},
+    AeadCore, AeadInPlace, Buffer, Error, NewAead,
+};
+use chacha20::ChaCha20Legacy;
+use cipher::NewCipher;
+use zeroize::Zeroize;
+
+use aead::generic_array::GenericArray;
+use cipher::{StreamCipher, StreamCipherSeek};
+use core::convert::TryInto;
+use poly1305::{
+    universal_hash::{NewUniversalHash, UniversalHash},
+    Poly1305,
+};
+use subtle::ConstantTimeEq;
+
+#[derive(Clone)]
+#[cfg_attr(docsrs, doc(cfg(feature = "legacy")))]
+pub struct ChaCha20Poly1305Legacy {
+    /// Secret key
+    key: Key,
+}
+
+impl NewAead for ChaCha20Poly1305Legacy {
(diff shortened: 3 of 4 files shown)


tarcieri:
Unfortunately it looks like you won't be able to use `Cipher` because the two constructions are slightly different in ways that I wasn't suspecting. My apologies for claiming this would be easy.

Here's the "djb" version:

https://github.com/jedisct1/libsodium/blob/6d566070b48efd2fa099bbe9822914455150aba9/src/libsodium/crypto_aead/chacha20poly1305/sodium/aead_chacha20poly1305.c#L37-L52

Here's the IETF version:

https://github.com/jedisct1/libsodium/blob/6d566070b48efd2fa099bbe9822914455150aba9/src/libsodium/crypto_aead/chacha20poly1305/sodium/aead_chacha20poly1305.c#L108-L127

If I'm reading these correctly, the Poly1305 inputs differ in the following way:

- djb version: `aad || len_bytes(aad) || msg || len_bytes(msg)`
- IETF version: `pad(msg) || pad(aad) || len_bits(aad) || len_bits(aad)`

Where `pad(...)` adds zero bytes until the input is aligned with the Poly1305 block size. You don't need to do that.

Here's an example of unpadded Poly1305:

https://github.com/RustCrypto/AEADs/blob/458a8a8/xsalsa20poly1305/src/lib.rs#L284

kamulos:
Thanks for the fast analysis, that helped a lot! I just hacked something into a commit, that seems to be working, but obviously is not any good.

My main problem is, that the `compute_unpadded` function you recommended consumes itself. I was not able to find anything for the `Poly1305`, that could be updated multiple times. Do you have a recommendation for that?

tarcieri:
@kamulos unfortunately the `poly1305` crate does not provide a buffered, non-block oriented API.

Sorry to stray even further from "this will be easy", but it seems like to implement this construction on top of the current API the `poly1305` crate provides, you'll need to add a buffering wrapper.

@newpavlov has written a number of abstractions for this sort of thing and may be able to help.

> My main problem is, that the compute_unpadded function you recommended consumes itself. I was not able to find anything for the Poly1305, that could be updated multiple times. Do you have a recommendation for that?

If you implement a block-sized buffer (16-bytes), you can pass data a block-at-a-time to Poly1305 using the [`UniversalHash::update`](https://docs.rs/universal-hash/0.4.0/universal_hash/trait.UniversalHash.html#tymethod.update) method, and then using [`Poly1305::compute_unpadded`](https://docs.rs/poly1305/0.7.0/poly1305/struct.Poly1305.html#method.compute_unpadded) to pass in any remaining data, regardless of whether or not it's aligned to the size of a block.

kamulos:
:smile: no worries, that doesn't sound too bad. I think I have all the necessary information now. I will prepare something as soon as I find time and send it to you for a review.

PR #308, opened 2021-05-07
Add aead/rand_core to default features to xsalsa20poly1305.

kurnevsky:
Currently it's not enabled by default even though rand_core dependency
is enabled by default.


tarcieri:
Thanks!

PR #309, opened 2021-05-08
Implement hash for PublicKey.

kurnevsky:
It's convenient to have PublicKey as key in a Map (for isntance, we [use it in tox](https://github.com/tox-rs/tox/blob/db7de02c6d699b03c014c8bc11d63101022cfbc4/tox_core/src/relay/client/connections.rs#L102)).


newpavlov:
@tarcieri 
I think we can merge it?

PR #311, opened 2021-05-15
Deoxys

zer0x64:
This is an initial implementation of the Deoxys AEAD. The Deoxys-II variant has been selected as the first choice for defense in-depth during the CAESAR competition.

A couple of things I checked:
- All the official test vectors passes (see the `tests/` folder)
- The crate is `[no_std]`
- The crate compiles and run in WASM.

There is still improvements to be done on the performance side, but this seems good enough for a first release.


deoxys/Cargo.toml
```diff
@@ -0,0 +1,34 @@
...
+
+[dependencies]
+aead = { version = "0.4", default-features = false }
+cipher = "0.3"
+subtle = "2.4.0"
+zeroize = { version = "1", default-features = false }
+
+[dev-dependencies]
+aead = { version = "0.4", features = ["dev"], default-features = false }
+hex = "0.4.3"
```
tarcieri, line 23: FWIW we generally use `hex-literal` (which allows hex consts that expand to byte arrays, and compile-time errors for malformed hex)

zer0x64: I replaced it everywhere!

tarcieri:
> Deoxys uses the AES round as the primitive for its operation, however there doesn't seem to be a way to use the aes crate for that purpose as the round functions aren't exposed. Right now the round function is reimplemented in 100% software in the aes_ref module, but this is slow, doesn't use AES-NI when possible and is probably full of side-channel attacks. I think the best person to ask here is @tarcieri ?

Unrelated to this PR, I was actually wondering about if exposing a more fine-grained API from the `aes` crate would be helpful the other day, especially as I'm currently [working on adding support for the ARMv8 cryptography extensions to the `aes` crate](https://github.com/RustCrypto/block-ciphers/pull/250).

Exposing the round function (as a very much "hazmat" API) in a way that's portable across `x86`/`x86-64` and `aarch64` targets would actually be pretty easy, or at least could be easy with some refactoring.

It's a bit trickier with the software implementation which uses a slightly modified SBox which omits some bitwise NOTs which are moved to the key schedule for efficiency reasons, however I think we could still build a common abstraction across all three.

Decryption is a bit trickier as different approaches are employed by the different backends (i.e. inverse cipher versus equivalent inverse cipher, as described in the Rijndael paper and FIPS 197), however these approaches are duals of each other and one can be used to implement the other.

I'd say it's something we should really address before merging.

zer0x64:
> I'd say it's something we should really address before merging.

Agreed! A lot of Deoxys strength relies on that it relies on the AES round function which already has fast and safe implementations, so duplicating the codebase for that part wouldn't be very efficient or safe. 

On another note, since this might take some time, do you think it would be a good idea to reserve the `deoxys` name under RustCrypto on crates.io?

tarcieri:
> On another note, since this might take some time, do you think it would be a good idea to reserve the deoxys name under RustCrypto on crates.io?

Snagged it 😉

tarcieri:
I opened https://github.com/RustCrypto/block-ciphers/issues/252 to track exposing the AES round function.

tarcieri:
@zer0x64 I think the trickiest thing about https://github.com/RustCrypto/block-ciphers/issues/252 will be handling decryption. Namely Intel's AES-NI and the ARMv8 AES intrinsics take somewhat different approaches:

https://en.wikipedia.org/wiki/Marina_Sirtis

I'd be curious to look at some other hardware-accelerated implementations of Deoxys and see what they do.

tarcieri:
@zer0x64 I think the trickiest thing about https://github.com/RustCrypto/block-ciphers/issues/252 will be handling decryption. Namely Intel's AES-NI and the ARMv8 AES intrinsics take somewhat different approaches:

https://blog.michaelbrase.com/2018/05/08/emulating-x86-aes-intrinsics-on-armv8-a/

I'd be curious to look at some other hardware-accelerated implementations of Deoxys and see what they do.

zer0x64:
The only hardware-accelerated Deoxys-II implementation I'm aware of is the Rust implementation by Oasis Labs(maybe their Go implementation is too), which only supports Deoxys-II on x86:

https://github.com/oasisprotocol/deoxysii-rust

Technically, Deoxys-II never requires a Deoxys-BC decryption because of the stream-cipher like design of the mode. However, I "think" that Deoxys-I actually requires decryption, but that's not something I can confirm without actually implementing it because that's the only surefire way to make sure I completely understand it.

Note that(for x86) only "AESENC" and maybe "AESDEC" are used in Deoxys. There is no "last" round like in AES, the key schedule is different and there is no need to directly call inverse MixCollumns.

Note that, once again, in the case of Deoxys, AddRoundKey can be either at the beggining or at the end of the round as it is the first and the last instruction of the entire sequence. Basically, you either do a manual AddRoundKey and then the rounds ending with AddRoundKey OR you do rounds that starts with AddRoundKey and do a manual AddRoundKey at the end.

tarcieri:
Sounds good.

Encryption-wise the ARMv8 instructions just split out Mix Columns/Inverse Mix Columns from the rest of the round function, and if you want control of when the round key XOR takes place, you can always pass a key of zeros to the hardware intrinsic.

So maybe a reasonable initial API would be one that looks like Intel's `AESENC`, i.e. the entire encryption round function, including Mix Columns, which accepts a round key and block to process as inputs.

zer0x64:
So as far as I can see, Deoxys CAN be done cleanly on intrinsics from both platforms (x86: XOR -> loop(AESENC), ARM: loop(AESE -> AESMC) -> XOR), however I don't really see a good way to abstract these into a single platform-independant method that can be used from Deoxys

zer0x64:
> Encryption-wise the ARMv8 instructions just split out Mix Columns/Inverse Mix Columns from the rest of the round function, and if you want control of when the round key XOR takes place, you can always pass a key of zeros to the hardware intrinsic.

I didn't think about the the null key!

zer0x64:
Deoxys has been designed with AES-NI in mind, and AES-NI uses the "Equivalent Inverse Cipher", so there must be a way to tweak the key schedule to work with it, although I might need to work a bit to get it to work. From the article, I also saw that both x86 and ARM AES instruction sets uses this approach, so there shouldn't be any problem on that side.

tarcieri:
>  I think all that's needed is to expose AESIMC, the ARM and the software equivalent(InverseMixCollumns) and apply the transformation to the round keys before using them, which should be pretty easy to implement on the Deoxys side.

So this is the tricky question for decryption: should the `aes` crate expose an equivalent inverse cipher API?

FIPS 197 (Appendix C.3) provides test vectors for:

- CIPHER (ENCRYPT)
- INVERSE CIPHER (DECRYPT)
- EQUIVALENT INVERSE CIPHER (DECRYPT)

Should we pick one of inverse cipher of equivalent inverse cipher to support? Or try to support both?

zer0x64:
I thinks I see the point: Should the AESIMC be done inside the `aes` crate or outside of it?

Personnally I think that exposing an "Inverse Cipher" API would be better for a couple of reasons. It feels more natural, AKA the encrypt and decrypt API would be the same, also it would not require to expose AESIMC to other crates, and AFAIK this wouldn't hurt performance as the intrinsics only supports Equivalent Inverse Cipher anyway and the inversion step would be done only once per round, regardless of which API we choose.

zer0x64:
Deoxys-I implemented and tested :tada: 

Quick question regarding the API:

Deoxys-II uses 15 bytes in the spec and internally in the cipher. However, 15-bytes nonces is a pretty non-standard size and even the the provided test vectors uses 16-bytes nonce, even if the least significant byte is never used. Should I make it so a 16 byte is required even though a byte isn't used or should I require a 15 byte nonce? (I think 15 would be better as it's a bit more flexible for the user and more efficient)

tarcieri:
Starting with a 15-byte nonce sounds good to me.

If 16-bytes is more helpful in practice (e.g. generic bounds) we can explore that later.

tarcieri:
I opened a preliminary PR with a raw round function API here, gated under a new `hazmat` feature: https://github.com/RustCrypto/block-ciphers/pull/257

It only supports the hardware intrinsic backends for now, but provides equivalent semantics between Intel AES-NI and the ARMv8 Cryptography Extensions, as tested against some of the FIPS 197 round function test vectors.

TODO: equivalent inverse cipher API

zer0x64:
I'll fix the clippy warnings and try to see if the exposed round functions does the trick.

zer0x64:
@tarcieri  I just got it working on my side, I also need the Inverse MixCollumns instruction to invert the keys before sending it to the round function on decryption. Right now it's working on my side but using the software implementation.

tarcieri:
Aah, ok. Maybe I should rename the module from `round` => `hazmat`.

As it were, there are hardware instructions for InvMixColumns on both x86 (`_mm_aesimc_si128`) and ARMv8 (`vaesimcq_u8`)

zer0x64:
I just pushed the changes to the round encryption/decryption to use the AES crate(although it's imported via a relative path so this will break everywhere expect my own setup for now). I added an `unsafe` block for now to use x86 AESIMC, but I'll change it to use the `aes` crate when available. I also made used a `cfg` gate to fallback on my software implementation.

tarcieri:
@zer0x64 if you update the `aes` crate there is now an `inv_mix_columns` function in the (newly renamed) `hazmat` module

pdugre:
Great! I'll do it when I can, and I will also update the PR so it pulls the `aes` crate from git instead of locally

zer0x64:
Great! I'll update it when I can and also update it so it pull the `aes` crate from git instead of locally

tarcieri:
@zer0x64 looks like it needs a rebase

zer0x64:
Done!

So I'll keep the `aes_ref` module for now as a software fallback and for the MixCollumn step(the one that doesn't have a dedicated x86 instruction). I'll eventually remove them when they will be available from the `aes` crate ( I'll open the issue after the merge).

What's left to do before removing the [WIP] tag is to cleanup `mode.rs` and add some benchmarks to see if the performance is good enough for a first release.

tarcieri:
> for the MixCollumn step(the one that doesn't have a dedicated x86 instruction)

With a clever hack it's possible to support Mix Columns with AES-NI: https://github.com/RustCrypto/block-ciphers/pull/260

zer0x64:
I just ran a benchmark and performance are horrendous, even with hardware acceleration. On my Ryzen 5950x, I'm getting anywhere from 50-250 cycles per bytes depending on the variant, compared to 1-3 for the other ciphers. My guess would be that the key schedule is at fault, since there is a major difference between 128 and 256 bit keys, so I'll investigate that after the `modes.rs` cleanup.

I'll commit the bench code as soon as it stops running!

zer0x64:
I profiled a bit and I can make the entire thing run ~10-40% faster simply by changing the LFSRs to lookup tables. However the reason I didn't initially do it was by fear of timing attacks. The official reference implementation use lookup tables for this.

Should I use lookup tables for the LFSRs for better performance? Or would it make the implementation more vulnerable to timing attacks?

tarcieri:
We've tried really hard to avoid LUTs elsewhere in the RustCrypto project.

Can you make a line note about the specific code in question? I'll take a look a bit later.

zer0x64:
> We've tried really hard to avoid LUTs elsewhere in the RustCrypto project.
> 
> Can you make a line note about the specific code in question? I'll take a look a bit later.

No problem, I'll add a TODO comment!

deoxys/src/deoxys_bc.rs
```diff
@@ -0,0 +1,173 @@
...
+// TODO: This operation is very slow
+// On Deoxys-II-256, shuffle_tweakey(), 
+//   which consists of h_substitution and lfsr2 and lfsr3, takes up 65% of the encryption time
+fn lfsr3(tk: &mut [u8]) {
+    for x in tk {
+        let feedback = (*x << 1) & 0x80;
+        *x = x.rotate_right(1);
+        *x ^= feedback;
+    }
+}
```
zer0x64, line 173: This is the part that's really slow

tarcieri: Oof, well my only thought there is a SIMD implementation, but doing a quick search I don't see any crates that make it easy to implement/synthesize that for LFSRs, which is a bit of a shame.

zer0x64:
As for the code cleanup, I'm not sure it's really worth it to try to reuse more code inside Deoxys-I as it might simply make the matter worse. I still have a nasty generics issue I'd like to fix (having to specify the Block Ciper twice in the variant type definition), but apart from that the rest will be more subtle changes when I come across them and will be re-done right before merging as the code can still get messed up quite a bit by the remaining changes. I can then consider that step "pretty much done".

That leaves the software fallback and optimisation. I don't think the `aes_ref` module is worth optimising as it will be completely removed as soon as the `aes` crate exposes it's software round and MixCollumns.

Should we wait for the software fallback of `aes` `hazmat` module before merging? Is there also a desired/required performance target we should try to reach before merging?

tarcieri:
I can take a look at backing the `hazmat` functions with the "soft" fixsliced implementation, but probably not until this weekend. I'd really like to get that working before cutting another release of the `aes` crate with that functionality, and I'd consider an `aes` release a prerequisite to this PR, as well as removing the `aes_ref` code.

As for things like performance, I'm not particularly worried so long as the implementation is correct.

zer0x64:
Great! I'll remove the WIP tag when I can get rid of `aes_ref`. In the meantime I'll work a bit more on code cleanup.

deoxys/src/deoxys_bc.rs
zer0x64: While I wouldn't say that's a good solution, I've played around with `unsafe` to interpret the whole input as a u128 and do all bytes at once and it seems to partially help (20% overall performance over the current implementation for the complete encryption/decryption process). Maybe there's way without `unsafe` code to hint the compiler to do it this way? 

```
// TODO: This operation is very slow
// On Deoxys-II-256, shuffle_tweakey(),
//   which consists of h_substitution and lfsr2 and lfsr3, takes up 65% of the encryption time
fn lfsr2(tk: &mut [u8]) {
    unsafe {
        let data = &mut *(tk as *mut [u8] as *mut u128);
        *data = ((*data << 1) & 0xFEFEFEFEFEFEFEFEFEFEFEFEFEFEFEFE) | (((*data >> 7) ^ (*data >> 5)) & 0x01010101010101010101010101010101);
    }
}

// TODO: This operation is very slow
// On Deoxys-II-256, shuffle_tweakey(),
//   which consists of h_substitution and lfsr2 and lfsr3, takes up 65% of the encryption time
fn lfsr3(tk: &mut [u8]) {
    unsafe {
        let data = &mut *(tk as *mut [u8] as *mut u128);
        *data = ((*data >> 1) & 0x7F7F7F7F7F7F7F7F7F7F7F7F7F7F7F7F) | (((*data << 7) ^ (*data << 1)) & 0x80808080808080808080808080808080);
    }
}
```

zer0x64: Another thing I just thought is that the actual input key(so the first 128 or 256 bits of the tweakey) and their multiple state for the key schedule can be precomputed as soon as the cipher is created and cached. This would basically mean that the LSFR and the h function would only be called once per round when the cipher is created

zer0x64:
Good news! :tada: 

I realised that the way the key schedule works, only the tweak part needs to be computer for each encryption/decryption of Deoxys-BC, and that the most expensive operations were only used on the key. That means that I could compute a huge part of the subkeys only in the `new()` method and reuse those values instead of re-computing them every time the tweak is changed. This results in 50%-150% better throughput(depending on the variant) and the cipher now reaches 40-100 cycles per bytes depending on the variant, as opposed to 50-250 before the optimisation.

deoxys/src/deoxys_bc.rs
zer0x64: Now vectorized correctly, closing.

deoxys/Cargo.toml
```diff
@@ -0,0 +1,36 @@
...
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+aead = { version = "0.4", default-features = false }
+subtle = "2.4.0"
+zeroize = { version = "1", default-features = false }
+
+[target.'cfg(target_feature="aes")'.dependencies]
+aes = { git = "https://github.com/RustCrypto/block-ciphers", rev="5e020f14d5fbff531bb6f9f15e9b97c366b8919c", features=["hazmat"]}
```
tarcieri, line 21: FYI, `aes` v0.7.3 is out with this feature/API.

Still no software fallback support, but I'll look at that when I get some time (or if anyone else would like to take a look, by all means, but it will be a bit tricky)

zer0x64: I've updated the import!

tarcieri:
Aside from the software fallback for AES, this is looking like a pretty reasonable spike at this point.

zer0x64:
I've removed `aes_ref` and ran the tests with both `target-feature=-aes` and `target-feature=+aes` (although I haven't tested with `force-soft`, but I might simply test it in WASM to test the software fallback). I guess what's left for a first release is simply to wait for a new `aes` version?

zer0x64:
I've tried to run it in WASM and I'm getting a bunch of unresolved modules from the `aes` crate. More specifically:
```
cpufeatures
aes_intrinsics
intrinsics
```

My guess is that those would also be unresolved on any non-x86/ARM platforms too

tarcieri:
Yep. I'm working on parallel encrypt/decrypt APIs in `hazmat`, then I'll cut a release.

zer0x64:
To reproduce, simply `cargo build --target wasm32-unknown-unknown`

tarcieri:
Aah, yeah the `hazmat` API is probably not gating things correctly yet.

I need to backfill CI to check it on platforms without CPU intrinsics (it previously had no software backend, so it wasn't possible)

tarcieri:
I just landed a few PRs in the `aes` crate, including https://github.com/RustCrypto/block-ciphers/pull/269 which adds parallel round APIs that operate over `aes::ParBlocks`, and https://github.com/RustCrypto/block-ciphers/pull/270 which should address the build failures.

Let me know if that all looks good and I can cut another release.

tarcieri:
Also feel free to mark as no-longer-draft / ready for review when you think it's ready. Looking pretty good now to me.

tarcieri:
FYI: I cut an `aes` v0.7.4 which includes software fallback and the parallel `cipher_round_par` and `equiv_inv_cipher_round_par` parallel APIs.

zer0x64:
Great! I'll check if everything's working on my side later today and how hard it is to make use of the parallel API, then I'll remove the WIP/draft tag!

zer0x64:
I've checked a bit and using the `aes` parallel block api will require a bit of work, mostly because of the way the memory is currently layed out for the round keys. Should I try to get it working before merging or this can be done later with another PR?

tarcieri:
Looking good now! I'll do a more detailed review when I have some time.

tarcieri:
Went ahead and merged. There are a few small things I might like to change, but I can do them as follow-ups when I have time.

PR #313, opened 2021-05-18
Expose aes' `armv8` feature in aes-{gcm,gcm-siv,siv}

aspenluxxxy:
I forgot if this needed, or if adding `aes = { version = "*", features = ["armv8"] }` will override the deps, If it's the latter, you can close this PR.

PR #318, opened 2021-05-31
aes-gcm(-siv): add nightly-only `armv8` feature

tarcieri:
This commit adds an `armv8` feature to the `aes-gcm`/`aes-gcm-siv` crates, which enables the corresponding `armv8` features on the `aes` crate and the `ghash`/`polyval` crates.

When enabled, these features use `stdarch`-based backends which leverage the ARMv8 Cryptography Extensions.

Presently `aarch64`-only.


tarcieri:
FYI @aspenluxxxy, changed my mind on #313 now that there are also `armv8` backends for `ghash` and `polyval`.

jrose-signal:
Can you tag aes-gcm-siv with this change as well? I'd love to delete our implementation of runtime-detected ARMv8 crypto support for aes-gcm-siv.

tarcieri:
Released in `aes-gcm-siv` v0.10.1

PR #321, opened 2021-06-03
AesPmacSiv features were not selectable by end users

maxburke:
This change exposes a feature to enable users to use AES-PMAC-SIV constructs.

Unfortunately it renamed the "pmac" feature because it conflicts with the pmac dependency.

### DIFF
--- a/aes-siv/Cargo.toml
+++ b/aes-siv/Cargo.toml
@@ -23,7 +23,7 @@ cmac = "0.6"
 crypto-mac = "0.11"
 ctr = "0.7"
 dbl = "0.3"
-pmac = "0.6"
+pmac = { version = "0.6", optional = true }
 zeroize = { version = "1", default-features = false }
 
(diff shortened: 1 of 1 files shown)


tarcieri:
> Unfortunately it renamed the "pmac" feature because it conflicts with the pmac dependency.

If you change the `pmac` dependency to be optional in Cargo.toml:

```
pmac = { version = "0.6", optional = true }
```

...then it will function as a `pmac` feature and you shouldn't need to change anything else. Not sure how/when it was made a hard dependency, but that's the core bug here.

maxburke:
Ah, neat, didn't know about that! Change made and re-pushed.

tarcieri:
Thank you! I'll try to cut another release soon.

Also note in general I'd recommend the `aes-gcm-siv` unless there's a particular reason you'd prefer PMAC (e.g. embedded device with AES accelerator).

PR #337, opened 2021-07-11
README: replace build badges column with MSRV

newpavlov:
I think MSRV information is more valuable than the build badges, especially considering that in this repository we have crates with 3 different MSRVs.

Also I've removed the `aes-ccm` mention, since the `ccm` crate should be preferred for now. Later we probably should introduce a folder with crates containing common common composition of primitives (AES-GCM, ChaCha20Poly1305, etc.).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,6 @@
 # RustCrypto: Authenticated Encryption with Associated Data (AEAD) Algorithms
 
 [![CodeCov Status][codecov-image]][codecov-link]
-![Rust Version][rustc-image]
 [![Project Chat][chat-image]][chat-link]
 [![dependency status][deps-image]][deps-link]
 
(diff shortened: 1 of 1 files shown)


tarcieri:
Alternatively we could get all of the MSRVs in sync, e.g. 1.51+.

That would make it possible to re-unify the `heapless` tests in the workflows, and set `resolver = "2"` for the whole repo.

newpavlov:
I don't think we should bump MSRV without a good reason. For example, at my current job we want to be able to build some crates with older toolchains (at the very least 1.41, since it's the version packaged in Debian).

And even assuming we had synchronized MSRVs, I feel that the build badges are somewhat pointless. I don't think they have been useful to us even once, despite them taking a significant space in our READMEs. Also we don't even have cron jobs in our workflows, so the only place where things can break is in PRs.

PR #348, opened 2021-07-19
Remove crate-local lock files

newpavlov:


tarcieri:
Weird, I guess those might've accidentally gotten checked in after running `cross` locally or something

newpavlov:
It's probably a leftover from the time of development, when those crates were not part of the workspace. At least I think it's the case for `ccm`.

PR #349, opened 2021-07-19
Pin zeroize dependency to v1.3 and subtle to v2.4

newpavlov:
`zeroize` in v1.4 has updated its MSRV to 1.51 breaking our MSRV promise. This PR pins its version to v1.3.x. `subtle` is also preemptively pinned to v2.4.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -28,7 +28,7 @@ dependencies = [
 
 [[package]]
 name = "aes-gcm"
-version = "0.9.2"
+version = "0.9.3"
 dependencies = [
  "aead",
  "aes",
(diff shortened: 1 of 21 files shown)


newpavlov:
@tarcieri 
Assuming you don't have objections, I can prepare PRs for other affected repositories.

tarcieri (approved):
@newpavlov sure, sounds good.

Just note that https://github.com/rustcrypto/elliptic-curves is MSRV 1.51 across the board and doesn't need it.

newpavlov:
The problem is that you may one day bump MSRV for `zeroize` v1.5, which would break MSRV build for the elliptic curve crates as well. So to be safe I think it's worth to pin 3rd party dependency versions for them as well.

tarcieri:
Yeah sure, it's OK to pin, just make sure to pin it to 1.4 instead of 1.3.

PR #355, opened 2021-08-24
Implement xchacha reduced-round variants

lain-dono:
XChaCha8Poly1305 and XChaCha12Poly1305 via ChaChaPoly1305 generalisation.


chacha20poly1305/Cargo.toml
```diff
@@ -28,13 +28,16 @@ zeroize = { version = "=1.3", default-features = false }
...
 alloc = ["aead/alloc"]
 heapless = ["aead/heapless"]
-reduced-round = ["chacha20"]
 stream = ["aead/stream"]
-xchacha20poly1305 = ["chacha20/xchacha"]
+xchacha20 = ["chacha20/xchacha"]
+xchacha20poly1305 = ["xchacha20"] # alias
+reduced-round = ["chacha20-reduced-round", "xchacha20-reduced-round"]
+chacha20-reduced-round = ["chacha20"]
+xchacha20-reduced-round = ["xchacha20"]
```
lain-dono, line 40: I left `xchacha20poly1305'. You may want to remove this in the future.

chacha20poly1305/src/lib.rs
```diff
@@ -150,24 +203,39 @@ pub type Tag = GenericArray<u8, U16>;
...
+/// XChaCha12Poly1305 (reduced round variant) Authenticated Encryption with Additional Data (AEAD).
+#[cfg(feature = "xchacha20-reduced-round")]
+#[cfg_attr(docsrs, doc(cfg(feature = "xchacha20-reduced-round")))]
+pub type XChaCha12Poly1305 = ChaChaPoly1305<XChaCha12, U24>;
 
 /// Generic ChaCha+Poly1305 Authenticated Encryption with Additional Data (AEAD) construction.
 ///
 /// See the [toplevel documentation](index.html) for a usage example.
-pub struct ChaChaPoly1305<C>
+pub struct ChaChaPoly1305<C, N: ArrayLength<u8> = U12>
```
lain-dono, line 236: The default option is used here, so this has no effect on compatibility.

chacha20poly1305/Cargo.toml
```diff
@@ -28,13 +28,16 @@ zeroize = { version = "=1.3", default-features = false }
...
 [features]
-default = ["alloc", "chacha20", "xchacha20poly1305"]
+default = ["alloc", "chacha20", "xchacha20"]
 std = ["aead/std", "alloc"]
 alloc = ["aead/alloc"]
 heapless = ["aead/heapless"]
-reduced-round = ["chacha20"]
 stream = ["aead/stream"]
-xchacha20poly1305 = ["chacha20/xchacha"]
+xchacha20 = ["chacha20/xchacha"]
```
tarcieri, line 36: I feel a bit of explanation is needed as to why there's a `chacha20` feature in the first place, as I feel there doesn't need to be an `xchacha20` feature.

This feature activates the `chacha20` *crate*.

The reason why it's optional is because the `c2-chacha` crate provided a faster implementation and also impl'd the RustCrypto traits, so people who really cared about performance could swap it in instead, and disable the `chacha20` crate.

With https://github.com/RustCrypto/stream-ciphers/pull/261 the gap has narrowed considerably (ty @str4d).

I think in the next breaking release we can make `chacha20` a mandatory dependency again, which would eliminate the need for this feature.

All that said, it seems like `xchacha20poly1305` and `xchacha20-reduced-round` could both activate `chacha20/xchacha`. And perhaps it would also make sense for simplicity's sake to get rid of `chacha20/xchacha` and always include the `XChaCha` family as the code size really isn't that big.

Anyway this is all a long winded way of saying it feels like there are too many features here and I take responsibility for at least part of that.

tarcieri:
Overall this looks great, although it seems the tests are failing due to unused import warnings under `--no-default-features`.

I also left one nit about a (cargo) "feature explosion" and am thinking it'd be good to simplify the overall feature set.

lain-dono:
I have tried to leave all changes within the PATCH change only. Not breaking change was the goal.

It also makes the documentation more beautiful. One of the rust features that we rarely notice, but love.

lain-dono:
It might make sense to remove the `reduced-round`. In that case, only `chacha20` and `xchacha20` will remain.
Of course, the aliases must remain in cargo.toml. Otherwise a new minor version is required.

tarcieri:
Ok, let's start with this, and I can circle back on slimming down features when I can make breaking changes to the `chacha20` crate in the next non-patch release.

tarcieri:
Thank you!

PR #365, opened 2021-08-29
chacha20poly1305: bump `chacha20` to v0.9; MSRV 1.51

tarcieri:
Bumps `chacha20` to the latest release, which has the following
additional changes:

- `chacha20` is now a hard dependency, now that the main use case for making it optional, using `c2-chacha` instead of `chacha20`, is irrelevant in that the performance difference has been eliminated.
- The `xchacha20` feature has been removed making the `XChaCha*` types always available.
- The `chacha20-reduced-round` and `xchacha20-reduced-round` features have been coalesced into a single `reduced-round` feature.

### DIFF
--- a/.github/workflows/chacha20poly1305.yml
+++ b/.github/workflows/chacha20poly1305.yml
@@ -22,7 +22,7 @@ jobs:
     strategy:
       matrix:
         rust:
-          - 1.49.0 # MSRV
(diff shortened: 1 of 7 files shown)


str4d:
Works for me: https://github.com/str4d/rage/pull/251

PR #367, opened 2021-08-29
chacha20poly1305 v0.9.0

tarcieri:
### Changed
- Bump `chacha20` to v0.9: now a hard dependency ([#365])
- MSRV 1.51+ ([#365])

### Removed
- `chacha20` feature: now a hard dependency ([#365])
- `xchacha20` feature: now always-on by default ([#365])
- `chacha20-reduced-round` and `xchacha20-reduced-round` have been coalesced into the `reduced-round` feature ([#365])

[#365]: https://github.com/RustCrypto/AEADs/pull/365


chacha20poly1305/CHANGELOG.md
```diff
@@ -4,6 +4,19 @@ All notable changes to this project will be documented in this file.
...
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 0.9.0 (2021-08-29)
+### Changed
+- Bump `chacha20` to v0.9: now a hard dependency ([#365])
+- MSRV 1.51+ ([#365])
+
+### Removed
+- `chacha20` feature: now a hard dependency ([#365])
+- `xchacha20` feature: now always-on by default ([#365])
```
str4d, line 14: "By default" implies it could be turned off, which I believe is not the case? It's just always-present.

tarcieri: Changed it to just say "now always-on"

PR #371, opened 2021-09-01
mgm: add target feature autodetection

newpavlov:


mgm/src/encdec.rs
```diff
@@ -0,0 +1,305 @@
...
+    C: BlockEncrypt,
+    C::BlockSize: MgmBlockSize,
+    E64: GfElement<N = U8>,
+    E128: GfElement<N = U16>,
+{
+    // ideally we would use type-level branching here
+    match C::BlockSize::USIZE {
+        8 => encrypt_inner::<C, E64>(args),
+        16 => encrypt_inner::<C, E128>(args),
+        _ => unreachable!(),
```
newpavlov, line 27: This part is somewhat ugly and breaks type-level safety of `encrypt_inner` and `decrypt_inner` functions (i.e. they may panic if incorrect `GfElement` will be provided to them), but AFAIK Rust currently does not have a nice way of doing type-level branching.

mgm/src/lib.rs
```diff
@@ -172,190 +116,100 @@ where
...
+    if #[cfg(all(
+        any(target_arch = "x86_64", target_arch = "x86"),
+        not(feature = "force-soft")
+    ))] {
+        #[target_feature(enable = "pclmulqdq")]
+        #[target_feature(enable = "ssse3")]
+        #[target_feature(enable = "sse2")]
+        unsafe fn wrapper<R: Sized>(f: impl FnOnce() -> R) -> R {
+            f()
+        }
```
newpavlov, line 154: I was not sure if this trick will work, but by inspecting generated assembly it looks like it does. Although I am not sure how "deep" can it work, e.g. I don't think it will influence code generation of items coming from different crates, unless they get inlined.

PR #373, opened 2021-10-25
xsalsa20poly1305: improve usage documentation

touilleMan:
As a new user of this library, I found surprisingly hard to understand how to use it 🤔

I think the documentation would benefit from providing an example closer to reality, hence I've changed the documentation to generate real key and nonce.

The drawback is this add the `rand` create to `doc-requirements`.

I wanted to to a similar change in `chacha20poly1305` (considering the two crates should be really similar), but they don't share the same API (at least `chacha20poly1305::generate_nonce` doesn't exist) is this expected ?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -357,10 +357,38 @@ dependencies = [
 ]
 
 [[package]]
+name = "ppv-lite86"
+version = "0.2.15"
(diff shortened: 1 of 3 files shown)


tarcieri:
This is a nice direction, but the API is changing and this PR is stale.

Please reopen if you'd like to continue it.

PR #401, opened 2022-03-16
Unpin `zeroize` version

trevyn:
Per https://github.com/RustCrypto/utils/issues/723#issuecomment-1068081155


poliorcetics:
To completely unblock the situation, a new release will be needed for all the crates here. Is this something done by maintainers only or can the version increase be done in a new commit on this branch by @trevyn ?

tarcieri:
We should bump the versions separately

PR #409, opened 2022-03-19
aes-gcm: bump `aes` crate to v0.8

tarcieri:


aes-gcm/Cargo.toml
```diff
@@ -32,8 +32,8 @@ hex-literal = "0.3"
 default    = ["aes", "alloc"]
 std        = ["aead/std", "alloc"]
 alloc      = ["aead/alloc"]
-armv8      = ["aes/armv8", "ghash/armv8"] # nightly-only
-force-soft = ["aes/force-soft", "ghash/force-soft"]
+armv8      = ["ghash/armv8"] # nightly-only
+force-soft = ["ghash/force-soft"]
```
tarcieri, line 36: It seems unfortunate there's no longer a way to transitively toggle on support for these across crates.

The `cfg`-based approach seems okay, but IMO there should be a single `cfg` which can force soft implementations across all crates, or force `nightly` armv8 support, rather than having to instruct people to pass two different `cfg`s for both `aes` and `ghash`

PR #432, opened 2022-06-27
ccm: update dependencies

rpiasetskyi:
```
* bump cipher and ctr dependencies to use it with the latest aes crate.
  Fix version difference.
* bump hex-literal.
* change MSRV to 1.56.0.
* remove separate job "heapless" in github workflow.
```

### DIFF
--- a/.github/workflows/ccm.yml
+++ b/.github/workflows/ccm.yml
@@ -22,7 +22,7 @@ jobs:
     strategy:
       matrix:
         rust:
-          - 1.41.0 # MSRV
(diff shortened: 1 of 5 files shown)


tarcieri (approved):
Looks good, thanks!

> should the version be updated from 0.4.4? If yes, to 0.4.5 or 0.5.0?

Make it 0.5.0-pre, thanks! We'll still want to update the `aead` crate to an as-yet-unreleased version 0.5.x before the final release.

PR #447, opened 2022-07-29
Make `ChaChaPoly1305` implement `ZeroizeOnDrop`

fjarri:
The zeroize-on-drop code was present before, but this PR announces that explicitly in the public API.

Some notes:
- To keep changes to the minimum, I did not add a `Zeroize` impl (probably not a good idea anyway)
- Had to restrict `zeroize` to `>=1.5`, hopefully that's ok.

### DIFF
--- a/chacha20poly1305/Cargo.toml
+++ b/chacha20poly1305/Cargo.toml
@@ -22,7 +22,7 @@ aead = { version = "0.5", default-features = false }
 chacha20 = { version = "0.9", features = ["zeroize"] }
 cipher = "0.4"
 poly1305 = "=0.8.0-pre.2"
-zeroize = { version = "1", default-features = false }
+zeroize = { version = "1.5", default-features = false }
 
 [dev-dependencies]
 aead = { version = "0.5", features = ["dev"], default-features = false }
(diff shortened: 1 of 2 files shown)


chacha20poly1305/Cargo.toml
```diff
@@ -22,7 +22,7 @@ aead = { version = "0.5", default-features = false }
 chacha20 = { version = "0.9", features = ["zeroize"] }
 cipher = "0.4"
 poly1305 = "=0.8.0-pre.2"
-zeroize = { version = "1", default-features = false }
+zeroize = { version = "1.5", default-features = false, features = ["derive"] }
```
tarcieri, line 25: This change adds 7 additional crate dependencies: `proc-macro2`, `quote`, `syn`, `synstructure`, `unicode-ident`, `unicode-xid`, `zeroize_derive`.

Furthermore, these are unavoidable hard dependencies.

For all of that, it's generating about 9 lines of code, 8 lines of which are the existing handwritten `Drop` impl.

I'd suggest leaving the handwritten impl as is and adding `impl ZeroizeOnDrop {}`. It's not worth adding that many dependencies for a single derive.

fjarri: Fair enough. I was uncertain about that too. Fixed.

tarcieri:
Thanks!

PR #465, opened 2022-08-08
chacha20poly1305: add `ChaChaPoly1305::generate_nonce`

tarcieri:
Adds a helper for generating random nonces, which is particularly useful with `XChaCha20Poly1305`.

Provides similar functionality to `XSalsa20Poly1305`.


tarcieri:
Perhaps this should be added to the `aead` crate, so we have a single place to provide guidance around random nonces

tarcieri:
It's now provided by `AeadCore` as of https://github.com/RustCrypto/traits/pull/1073

PR #469, opened 2022-08-16
Sundae

Schmid7k:
This is an optimized implementation of the (SUNDAE)[https://tosc.iacr.org/index.php/ToSC/article/view/7296/6470] AEAD cipher using x86_64 instructions. SUNDAE reached the second stage of the (NIST lightweight cryptography competition)[https://csrc.nist.gov/Projects/lightweight-cryptography] as part of SUNDAE-GIFT and is generally paired well with low area, low power consumption block ciphers (like GIFT), because it was specifically designed for resource constrained environments such as IoT devices. It can also be used as a MAC algorithm, if only associated data is provided.

This pull request contains a full implementation of the mode, no features missing (at least to my knowledge), though it would definitely be nice to implement an ARM version in the future.

It comes with a pre-defined SundaeAes type using Aes128 as underlying block cipher, though I think I would swap that out for GIFT once (#322)[https://github.com/RustCrypto/block-ciphers/pull/322] is approved. 

I was able to benchmark an optimized C vs this implementation and found, that both operate almost equally, with C being around 0.4 cpb faster for lower buffer sizes and 0.12 cpb for bigger buffer sizes. On an Intel Core i7 8700k with 3.7GHz core clock the Rust implementation performed at 5.595 - 5.387cpb and the C implementation at 5.18 - 5.265cpb for buffer sizes from 1KiB to 16KiB, both using Aes128 as block cipher.

The crate is [no_std] but one thing I have to point out is that I am using the `u8x16` type from `core::simd`, which requires me to enable `#![feature(portable_simd)]`, because it is still considered an unstable library feature.


Schmid7k:
And a rebase for SUNDAE as well.

Issue #471, opened 2022-08-21
Key from slice breaks in aes-gcm-0.10.1


Ontokrat:
Hello,

The following code in `aes-gcm-0.9.4`, to generate a cipher from a known key, now breaks in `aes-gcm-0.10.1`.
>let cipher = Aes256Gcm::new(Key::from_slice(key.as_ref()));

`key` is a byte string, like `b"a key"`.

The error is:
>type annotations needed: cannot satisfy `<_ as aes_gcm::KeySizeUser>::KeySize == aes_gcm::aead::generic_array::typenum::UInt<aes_gcm::aead::generic_array::typenum::UInt<aes_gcm::aead::generic_array::typenum::UInt<aes_gcm::aead::generic_array::typenum::UInt<aes_gcm::aead::generic_array::typenum::UInt<aes_gcm::aead::generic_array::typenum::UInt<aes_gcm::aead::generic_array::typenum::UTerm, aes_gcm::aead::consts::B1>, aes_gcm::aead::consts::B0>, aes_gcm::aead::consts::B0>, aes_gcm::aead::consts::B0>, aes_gcm::aead::consts::B0>, aes_gcm::aead::consts::B0>`
cannot satisfy `<_ as aes_gcm::KeySizeUser>::KeySize == aes_gcm::aead::generic_array::typenum::UInt<aes_gcm::aead::generic_array::typenum::UInt<aes_gcm::aead::generic_array::typenum::UInt<aes_gcm::aead::generic_array::typenum::UInt<aes_gcm::aead::generic_array::typenum::UInt<aes_gcm::aead::generic_array::typenum::UInt<aes_gcm::aead::generic_array::typenum::UTerm, aes_gcm::aead::consts::B1>, aes_gcm::aead::consts::B0>, aes_gcm::aead::consts::B0>, aes_gcm::aead::consts::B0>, aes_gcm::aead::consts::B0>, aes_gcm::aead::consts::B0>`

The doc in `aes-gcm-0.9.4`
>let key = Key::from_slice(b"an example very very secret key.");
let cipher = Aes256Gcm::new(key);

changed in `aes-gcm-0.10.1` for
>let key = Aes256Gcm::generate_key(&mut OsRng);
let cipher = Aes256Gcm::new(&key);

How in `aes-gcm-0.10.1` generate a key from a known byte string?

Thanks.

tarcieri:
`Key` is generic around the cipher, and the compiler can’t infer the generic parameter.

[...]

PR #472, opened 2022-08-21
Example in doc of key from known string - aes-gcm

Ontokrat:
From #471

### DIFF
--- a/aes-gcm/src/lib.rs
+++ b/aes-gcm/src/lib.rs
@@ -20,7 +20,7 @@
 //!     Aes256Gcm, Nonce // Or `Aes128Gcm`
 //! };
 //!
-//! let key = Aes256Gcm::generate_key(&mut OsRng);
+//! let key = Aes256Gcm::generate_key(&mut OsRng); // Or let key = Key::<Aes256Gcm>::from_slice(b"an example very very secret key.");
 //! let cipher = Aes256Gcm::new(&key);
 //! let nonce = Aes256Gcm::generate_nonce(&mut OsRng); // 96-bits; unique per message
 //! let ciphertext = cipher.encrypt(&nonce, b"plaintext message".as_ref())?;


percy507:
This is really helpful.

tarcieri:
This is an example from the old documentation I replaced showing key generation.

In hindsight, I think it's bad to recommend this API, since it accepts a slice and panics if it's the wrong length.

aes-gcm/src/lib.rs
```diff
@@ -20,7 +20,7 @@
 //!     Aes256Gcm, Nonce // Or `Aes128Gcm`
 //! };
 //!
-//! let key = Aes256Gcm::generate_key(&mut OsRng);
+//! let key = Aes256Gcm::generate_key(&mut OsRng); // Or let key = Key::<Aes256Gcm>::from_slice(b"an example very very secret key.");
```
tarcieri, line 23: Some issues wit hthis:

1. Having it on the same (very long) line like this will make it hard to see
2. The example should use a non-panicking conversion
3. Having it in a comment means it won't get doctested

percy507:
> In hindsight, I think it's bad to recommend this API, since it accepts a slice and panics if it's the wrong length.

You are right. I have experience that. 😂

But is there any recommend way for using custom key instead of the generated one?

tarcieri:
There's a way to do a non-panicking `TryFrom` conversion but it's a bit gnarly:

```rust
let key = Key::<Aes256Gcm>::from(<[u8; 32]>::try_from(key_slice)?);
```

newpavlov:
@tarcieri 
It should be sufficient to transform key into `[u8; 32]` using `TryInto` and use `key.into()` in places where the key is used.

tarcieri:
@newpavlov for the purposes of documentation I think it makes a lot more sense to document `Aes256Gcm::new_from_slice` than a `GenericArray` conversion

Mercanuis:
Are there any updates or a PR for the documentation on how to use `Aes256Gcm::new_from_slice`?

newpavlov:
Closing in favor of #493.

PR #476, opened 2022-09-02
ccm: Mention AES-CCM in README

chrysn:
This adds a mention of AES-CCM in CCM's README as an example.

AES-CCM is the most prominent use case of CCM, as evidenced by the wikipedia articles both of CCM and its users always listing that combination.

Practically, this ensures that users looking for `aes-ccm` at crates.io find this, especially if https://github.com/martindisch/aes-ccm/issues/9 is not followed up on. (Concrete example: I was a few steps short of starting to update the aes-ccm crate to recent traits when I found that issue which pointed me here; the present fix would have spared me that detour).

### DIFF
--- a/ccm/README.md
+++ b/ccm/README.md
@@ -11,6 +11,9 @@ Pure Rust implementation of the Counter with CBC-MAC ([CCM]) mode ([RFC 3610]): 
 Authenticated Encryption with Associated Data ([AEAD]) algorithm generic over
 block ciphers with block size equal to 128 bits.
 
+For example, it can be combined with AES into the various parametrizations of
+AES-CCM.
+
 [Documentation][docs-link]
 
 ## Security Notes


newpavlov:
Hm, I would think looking at the [crate docs](https://docs.rs/ccm/) would have been sufficient. Have you seen it?

I think a better solution could be to copy the docs content into the readme. Later on MSRV bump we could use `#! [doc = include_str! ("README.md")]` to remove duplication.

chrysn:
On Fri, Sep 02, 2022 at 05:37:55AM -0700, Artyom Pavlov wrote:
> Hm, I would think looking at the [crate docs](https://docs.rs/ccm/)
> would have been sufficient. Have you seen it?

The crate docs are good and helpful, but one needs to find that the
`ccm` crate can provide this when looking for an `aes-ccm` crate (like
there is `aes-gcm`) before one gets to the documentation.

> I think a better solution could be to copy the docs content into the
> readme. Later on MSRV bump we could use `#! [doc = include_str!
> ("README.md")]` to remove duplication.

That'd be a viable option if the string "AES-CCM" showed up there (but
that should be doable easily).

newpavlov:
We should make `aes-ccm` a simple wrapper around `aes` and `ccm`, and keep it updated. The owner has transferred access to us, but we somewhat forgot about that.

>That'd be a viable option if the string "AES-CCM" showed up there

Yeah, you can add something like this:
>Simple usage example (allocating, no associated data) which uses block cipher implementation from the `aes` crate to form AES-CCM:

tarcieri:
I think this is fine.

PR #493, opened 2022-11-28
aes-gcm: improve doc example

newpavlov:
Alternative to #472

### DIFF
--- a/aes-gcm/src/lib.rs
+++ b/aes-gcm/src/lib.rs
@@ -12,15 +12,33 @@
 //!
 //! Simple usage (allocating, no associated data):
 //!
-#![cfg_attr(all(feature = "getrandom", feature = "std"), doc = "```")]
-#![cfg_attr(not(all(feature = "getrandom", feature = "std")), doc = "```ignore")]
-//! # fn main() -> Result<(), Box<dyn std::error::Error>> {
+//! ```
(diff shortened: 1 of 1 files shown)


tarcieri:
I did want to leave some comments here but didn't get around to it yet.

The way of avoiding a `std` dependency was clever, however I think the example contains a lot of information about how to convert key types which distracts from the actual encryption workflow. I thought it might be worth splitting into two sections, one on the actual encryption workflow keeping it as straightforward as possible, and another on how to do key type conversions.

Also note this documentation is something of boilerplate which is repeated across the various crates.

newpavlov:
I do not plan to cut any releases in the near future, so there is a plenty of time to refine the docs further.

>I thought it might be worth splitting into two sections, one on the actual encryption workflow keeping it as straightforward as possible, and another on how to do key type conversions

The `aead` crate docs could be a good place to put such docs into, but we again encounter the cyclical dependency issue.

tarcieri:
I think it's nice to have documentation specialized to using each crate, although I'm not sure of a better way other than templating to make cross-cutting changes across all crates in the repo.

With README.md-based documentation, at least that'd be a little easier, as we'd only have to edit/template Markdown as opposed to Markdown-inside-rustdoc-comments

PR #495, opened 2022-12-13
Add correct link to Too much crypto

skaunov:
Currently at https://docs.rs/chacha20poly1305/latest/chacha20poly1305/#supported-algorithms it renders link to audit from <https://github.com/RustCrypto/AEADs/blob/57956bf2d8273e7b3fddd127aa7116ad5ef1c5ad/chacha20poly1305/README.md?plain=1#L82>.


tarcieri:
Can you run rustfmt?

tarcieri:
Seems to be some unrelated removal of absolute import paths which is breaking the tests… do you have some nonstandard rustfmt config which applied when you ran it?

You need to put those imports back the way they were.

skaunov:
Frankly, never used `rustfmt` before, so made it in most naive way, and didn't expect it could break things. Will try to config it right.

skaunov:
LOL. I literally ran `rustfmt`. Yeah, `cargo fmt` just amends the line break. Thanks for the hint!

Issue #500, opened 2023-01-12
aes-gcm partial tag


ALaggyDev:
Hi,
I am trying to use the `gcm` crate but with a 12 byte authorization tag, instead of the standard 16 byte authorization tag.
I can't really decrypt data because the program only know the first 12 byte tag.

According to online documentation, tag size can be customized in aes/gcm (by using the leftmost bytes).
So I am wondering if there is a possibility of implementing custom tag size in the `gcm` crate.
Thanks!

tarcieri:
We can make tag length a generic parameter like nonce size is, although that will be a breaking change.

Per [NIST SP 800-38D](https://csrc.nist.gov/publications/detail/sp/800-38d/final) Appendix C, the only allowed sizes are 96, 104, 112, 120 or 128, so we'll need a marker trait for the tag size to bless those particular sizes as acceptable.

ALaggyDev:
Is making tag lengh generic parameter really a breaking change if we set a default generic type?

Can something like this work?
```rust
/// AES-GCM tags.
pub type Tag<TagSize = U16> = GenericArray<u8, TagSize>;

pub struct AesGcm<Aes, NonceSize, T = U16>
where T: TagSize { ... }

// Implement for U8, U12, U13, U14, U15, U16
pub trait TagSize {}
```

Every new generic parameter have a default type, so I think it's fine?

tarcieri:
IIRC we tried that with `NonceSize` and ran into problems, although I don't remember what they were offhand.

Feel free to attempt it and open a PR (although if you do, please use something more descriptive than `T` as the generic parameter name)

dmitry-zakablukov:
@tarcieri , can you please publish a new version of `aes-gcm` package with variable TagSize? I really need these changes in my code.

tarcieri:
Done: #530

dmitry-zakablukov:
@tarcieri , thank you!

PR #501, opened 2023-01-12
gcm: partial tag size

ALaggyDev:
Fix of #500

Shouldn't be a breaking change but I am not sure exactly.


aes-gcm/src/lib.rs
```diff
@@ -129,7 +131,14 @@ pub const C_MAX: u64 = (1 << 36) + 16;
 pub type Nonce<NonceSize> = GenericArray<u8, NonceSize>;
 
 /// AES-GCM tags.
-pub type Tag = GenericArray<u8, U16>;
+pub type Tag<TagSize = U16> = GenericArray<u8, TagSize>;
+
+/// Trait implemented for valid tag sizes, i.e.
+/// [`U12`][consts::U12], [`U13`][consts::U13], [`U14`][consts::U14],
+/// [`U15`][consts::U15] and [`U16`][consts::U16].
+pub trait TagSizeLen: private::SealedTag {}
```
tarcieri, line 139: To bikeshed some names:

```suggestion
pub trait TagSize: private::SealedTagSize {}
```

aes-gcm/src/lib.rs
```diff
@@ -170,27 +198,33 @@ pub struct AesGcm<Aes, NonceSize> {
...
+
+    /// Length of the tag.
+    tag_size: PhantomData<TagSize>,
 }
 
-impl<Aes, NonceSize> KeySizeUser for AesGcm<Aes, NonceSize>
+impl<Aes, NonceSize, TagSize> KeySizeUser for AesGcm<Aes, NonceSize, TagSize>
 where
     Aes: KeySizeUser,
+    TagSize: self::TagSize,
```
tarcieri, line 209: Kind of surprised these pass the `unused_qualifier` lint. Seems like they should just be:

```suggestion
    TagSize: TagSize,
```

ALaggyDev: Yeah I added the `self::` because the generic type `TagSize` and the trait `TagSize` have name collisions.
So I had to add the `self::`, unless you want to change their names of course.

tarcieri: Removing `self::` should be fine?

Unless you’re saying it’s confusing?

We can hopefully change them into const generic parameters after migrating to `hybrid-array` in the next release

ALaggyDev: The compiler throws a compile error after removing `self::`.

```
error[E0404]: expected trait, found type parameter `TagSize`
   --> aes-gcm\src\lib.rs:191:14
    |
137 | pub trait TagSize: private::SealedTagSize {}
    | -------------------------------------------- you might have meant to refer to this trait
...
189 | pub struct AesGcm<Aes, NonceSize, TagSize = U16>
    |                                   ------- found this type parameter
190 | where
191 |     TagSize: TagSize,
    |              ^^^^^^^ not a trait
```

tarcieri: Aah, ok

tarcieri:
Thanks!

PR #506, opened 2023-02-28
Fix typo & Refine code style

zkonge:


newpavlov:
Closing in favor of #507 which uses a slightly better approach.

PR #514, opened 2023-03-17
Import `ascon-aead`

sebastinas:
The next step for completing https://github.com/sebastinas/ascon-aead/issues/5.


ascon-aead/Cargo.toml
```diff
@@ -0,0 +1,38 @@
...
+[dependencies]
+aead = { version = "0.5", default-features = false }
+subtle = { version = "2", default-features = false }
+zeroize = { version = "1.5", optional = true, default-features = false, features = [
+    "derive",
+] }
+ascon = "0.3"
+
+[dev-dependencies]
+hex = "0.4"
```
tarcieri, line 26: FWIW we usually use `hex-literal` for this (which parses hex at compile time and emits `[u8; N]` arrays), but I'd be happy to switch to that in a followup PR

ascon-aead/LICENSE-MIT
```diff
@@ -0,0 +1,25 @@
+Copyright (c) 2019 The RustCrypto Project Developers
```
tarcieri, line 1: Maybe update the copyright year here

ascon-aead/src/asconcore.rs
```diff
@@ -0,0 +1,379 @@
...
+            self.state.permute_6();
+        } else {
+            self.state.permute_8();
+        }
+    }
+
+    fn process_associated_data(&mut self, mut associated_data: &[u8]) {
+        if !associated_data.is_empty() {
+            // TODO: rewrite with as_chunks once stabilized
+            // https://github.com/rust-lang/rust/issues/74985
```
tarcieri, line 222: Absent `as_chunks` it still seems like you could use `chunks_exact`?

ascon-aead/src/asconcore.rs
```diff
@@ -0,0 +1,379 @@
...
+impl Parameters for Parameters80pq {
+    type KeySize = U20;
+    type InternalKey = InternalKey20;
+
+    const COUNT: usize = 8;
+    const IV: u64 = 0xa0400c0600000000;
+}
+
+/// Core implementation of Ascon for one encryption/decryption operation
+pub(crate) struct AEADCore<'a, P: Parameters> {
```
tarcieri, line 170: I think this should be `AeadCore` per https://rust-lang.github.io/api-guidelines/naming.html

> In UpperCamelCase, acronyms and contractions of compound words count as one word: use `Uuid` rather than `UUID`

ascon-aead/src/asconcore.rs
```diff
@@ -0,0 +1,379 @@
...
+        } else {
+            0
+        };
+        self.state[sidx] ^= pad(ciphertext.len());
+        if !ciphertext.is_empty() {
+            let cx = u64_from_be_bytes_partial(ciphertext);
+            self.state[sidx] ^= cx;
+            ciphertext.copy_from_slice(&u64::to_be_bytes(self.state[sidx])[0..ciphertext.len()]);
+            self.state[sidx] = clear(self.state[sidx], ciphertext.len()) ^ cx;
+        }
```
tarcieri, line 313: This seems like a candidate for `chunks_exact_mut`

ascon-aead/src/asconcore.rs
```diff
@@ -0,0 +1,379 @@
...
+            message = &mut message[8..];
+            1
+        } else {
+            0
+        };
+        self.state[sidx] ^= pad(message.len());
+        if !message.is_empty() {
+            self.state[sidx] ^= u64_from_be_bytes_partial(message);
+            message.copy_from_slice(&u64::to_be_bytes(self.state[sidx])[0..message.len()]);
+        }
```
tarcieri, line 279: Another `chunks_exact_mut` candidate

ascon-aead/src/lib.rs
```diff
@@ -0,0 +1,290 @@
+// Copyright 2021-2022 Sebastian Ramacher
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+#![no_std]
+#![cfg_attr(docsrs, feature(doc_cfg))]
```
tarcieri, line 5: We're migrating to `doc_auto_cfg` FWIW (not sure it's really relevant here either way)

ascon-aead/src/asconcore.rs
sebastinas: Renamed to `AsconCore`.

ascon-aead/Cargo.toml
sebastinas: The test vectors are currently read during runtime from the data files. A switch to `hex-literal` would require a rewrite of the current tests.

tarcieri: @newpavlov `hex-literal` feature request: `include_hex!` which works like `include_bytes!`/`include_str!`

ascon-aead/src/asconcore.rs
sebastinas: Yes. Thanks for the hint.

ascon-aead/src/asconcore.rs
sebastinas: Changed to use `chunks_exact_mut`.

ascon-aead/src/asconcore.rs
sebastinas: Also done.

ascon-aead/src/lib.rs
sebastinas: Fixed.

tarcieri (approved):

tarcieri:
@sebastinas I've sent you an invite to the AEADs team.

After you've accepted it, can you do:

```
$ cargo owner -a github:rustcrypto:aeads
```

...to allow the AEADs team to publish `ascon-aead`?

sebastinas:
Thanks!

```
$ cargo owner -a github:rustcrypto:aeads  
    Updating crates.io index
       Owner team github:rustcrypto:aeads has been added as an owner of crate ascon-aead
```

ascon-aead/Cargo.toml
newpavlov: @tarcieri
Could you create an issue for it?

BTW I don't think it would've helped here, since the test vector files contain non-hex data. I think a better approach would be to convert them to the blobby format and use the `aead::new_test!` macro`.

sebastinas: In the long run I think it would make more sense to support the format of the test vector files. The format is the same (except for the provided fields) is the same for all current standardization processes run by NIST. The test vectors for all the post-quantum signatures and KEMs look the same.

newpavlov: The test vectors get included into published packages, so it's worth to minimize them. Even with gzip, binary encoding produces a smaller result.

tarcieri: @sebastinas for what it's worth, we have an extractor for a similar use case: extracting files from Wycheproof, so we can convert to a common format:

https://github.com/RustCrypto/utils/tree/master/wycheproof2blb

@newpavlov the minute size wins for "blobby" files over e.g. git's own usage of gzip goes away the instant we need to modify any of the files and git has to duplicate the entire contents of the file because it treats binary data as opaque git objects.

It also removes the ability to add any kind of context to test vectors, like their provenance.

newpavlov: For similar test vector files I was using a quick Python script which generates file with only hex data. it then gets converted by the utility in the `blobby` crate. I may do conversion for this crate myself a bit later.

@tarcieri 
I was talking primarily about size of published packages. Not only they get downloaded much more often than repository gets cloned (outside of CI jobs), but we include them with each release.

In practice, we modify `blb` files quite rarely (I can remember only 2-3 cases). Usually, we add new test vectors as separate files even if we test the same algorithm (e.g. we did so with the Wycheproof vectors).

Provenance is better documented not in test vector files, but in documentation of tests which use those files. It's the first place developers will look into when inspecting crate tests. For example, the test files in this PR do not include any provenance information, so we have to document it separately either way.

tarcieri: Any time there's been a test failure in the Wycheproof-generated files it's been a massive pain to debug, because while the original files contain detailed information about the test case in question. I have to manually go from the blobby test failure back to the original test vectors to get that information back, which is quite confusing.

In general having a bespoke, opaque binary format without great tooling to assist working with it is an impediment to every new project contributor.

PR #515, opened 2023-03-18
benches: only use criterion CPB measurements on x86_64 and x86 targets

brxken128:
This also bumps the versions of `criterion`, `aes` and `criterion-cycles-per-byte`.

I'm not sure how verbose I should've been with the `cfg`s, let me know if there's a preferred style and I'd be happy to change them!


tarcieri:
That looks fine, thanks!

PR #518, opened 2023-03-21
benches: fix compilation

brxken128:
I must apologize - it appears my PR from the other day (#515) broke benchmarks on `x86_64` and `x86`.

I didn't realize as I was developing on `Aarch64`, but I'm not too sure why CI didn't pick it up.

The benchmark crate also refused to compile due to the `ascon-aead` benches - they attempt to use `rand` as a dependency but it wasn't specified.


tarcieri:
Weird, wonder why CI isn't catching this?

PR #527, opened 2023-05-19
ascon-aead: Add ascon sponge Zeroize feature

aewag:
Not zeroizing the state may expose the private key.

As the MSRV is below 1.60 the `dep:` syntax is not available. Hence, the second feature `sponge-zeroize`. Any other suggestions to handle it better are welcome.

This relies on https://github.com/RustCrypto/sponges/pull/57 and a new release of the `ascon` crate.

### DIFF
--- a/ascon-aead/Cargo.toml
+++ b/ascon-aead/Cargo.toml
@@ -17,7 +17,7 @@ rust-version = "1.56"
 [dependencies]
 aead = { version = "0.5", default-features = false }
 subtle = { version = "2", default-features = false }
-zeroize = { version = "1.6", optional = true, default-features = false, features = [
+zeroize_crate = { package = "zeroize", version = "1.6", optional = true, default-features = false, features = [
     "derive",
 ] }
 ascon = "0.4"
@@ -27,7 +27,7 @@ hex-literal = "0.3"
 aead = { version = "0.5", features = ["alloc"] }
 
 [features]
-default = ["alloc", "getrandom", "zeroize"]
+default = ["alloc", "getrandom"]
 std = ["aead/std", "alloc"]
 alloc = ["aead/alloc"]
 arrayvec = ["aead/arrayvec"]
@@ -35,6 +35,7 @@ getrandom = ["aead/getrandom", "rand_core"]
 heapless = ["aead/heapless"]
 rand_core = ["aead/rand_core"]
 stream = ["aead/stream"]
+zeroize = ["zeroize_crate", "ascon/zeroize"]
 
 [package.metadata.docs.rs]
 all-features = true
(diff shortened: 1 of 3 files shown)


ascon-aead/Cargo.toml
```diff
@@ -20,22 +20,23 @@ subtle = { version = "2", default-features = false }
 zeroize = { version = "1.6", optional = true, default-features = false, features = [
     "derive",
 ] }
-ascon = "0.3"
+ascon = { git = "https://github.com/aewag/sponges", branch = "add-zeroize-ascon", default-features = false }
```
sebastinas, line 23: ascon 0.4.0 is now available.

aewag: I just updated it.

ascon-aead/Cargo.toml
```diff
@@ -20,22 +20,23 @@ subtle = { version = "2", default-features = false }
...
-default = ["alloc", "getrandom", "zeroize"]
+default = ["alloc", "getrandom", "zeroize", "sponge-zeroize"]
 std = ["aead/std", "alloc"]
 alloc = ["aead/alloc"]
 arrayvec = ["aead/arrayvec"]
 getrandom = ["aead/getrandom", "rand_core"]
 heapless = ["aead/heapless"]
 rand_core = ["aead/rand_core"]
 stream = ["aead/stream"]
+sponge-zeroize = ["ascon/zeroize"]
```
newpavlov, line 39: I think you can simply use `zeroize ="ascon/zeroize"]`.

ascon-aead/Cargo.toml
```diff
@@ -20,22 +20,23 @@ subtle = { version = "2", default-features = false }
...
+ascon = "0.4"
 
 [dev-dependencies]
 hex = "0.4"
 aead = { version = "0.5", features = ["alloc"] }
 spectral = { version = "0.6", default-features = false }
 
 [features]
-default = ["alloc", "getrandom", "zeroize"]
+default = ["alloc", "getrandom", "zeroize", "sponge-zeroize"]
```
newpavlov, line 31: As noted in the other PR, usually, we do not enable `zeroize` by default.

ascon-aead/Cargo.toml
aewag: As `ascon-aead` has an optional `zeroize` dependency, there is a naming conflict.

`dep:` would solve this, but is only available with 1.60 or later. The current MSRV of `ascon-aead` is 1.56.

ascon-aead/Cargo.toml
aewag: Both features, `zeroize` and `sponge-zeroize`, are now removed

ascon-aead/Cargo.toml
tarcieri: One option is to rename the zeroize crate with a `package` attribute, e.g.:

```rust
zeroize_crate = { package = "zeroize", ... }
```

Alternatively we could bump MSRV/version, although that'd be easier to do to the entire repo at the same time.

sebastinas: I'd prefer the first option and delay the change to the MSRV until it's performed for the entire repo.

aewag: Thanks for the suggestion :+1: I updated it accordingly

PR #532, opened 2023-05-26
fix: `missing_debug_implementations`

JonathanWoollett-Light:
Sets `missing_debug_implementations` to warn and adds missing `std::fmt::Debug` implementations.

When tracing/logging/debugging a program it is common to debug print inputs and outputs of function, to do this generally dependencies need to offer `std::fmt::Debug` implementations.

I won't repeat what is written under [`missing_debug_implementations`](https://doc.rust-lang.org/rustc/lints/listing/allowed-by-default.html#missing-debug-implementations) which gives some more explanation.


JonathanWoollett-Light:
Blocked on https://github.com/RustCrypto/block-modes/pull/46

newpavlov:
This PR got outdated, so I will close it in favor of tracking issue #599.

PR #543, opened 2023-07-22
xsalsa20poly1305: remove source code

tarcieri:
The successor is the `crypto_secretbox` crate:

https://github.com/RustCrypto/nacl-compat/tree/master/crypto_secretbox


jrose-signal:
If I look at [the docs for `crypto_secretbox`](https://docs.rs/crypto_secretbox/0.1.1/crypto_secretbox/index.html) it still suggests to use xsalsa20poly1305:

> This algorithm has largely been replaced by the newer IETF variant of [ChaCha20Poly1305](https://github.com/RustCrypto/AEADs/tree/master/chacha20poly1305) (and the associated [XChaCha20Poly1305](https://docs.rs/chacha20poly1305/latest/chacha20poly1305/struct.XChaCha20Poly1305.html)) AEAD ciphers ([RFC 8439](https://tools.ietf.org/html/rfc8439)), but is useful for interoperability with legacy NaCl-based protocols.

I could ask what changed, but that's probably not important, so I'll just stick with "is the current release up to date other than that"?

tarcieri:
Yeah, `xsalsa20poly1305` is retired, replaced by `crypto_secretbox` which supports both XSalsa20Poly1305 and libsodium-flavored XChaCha20Poly1305.

I could potentially add the CHANGELOG.md back but I'd really like to just completely delete the `xsalsa20poly1305` directory at some point.

PR #550, opened 2023-09-16
Import `offset-cookbook-mode`

sgmenda:
Re https://github.com/sgmenda/offset-cookbook-mode/issues/1

I slightly cleaned up the code, made the implementation generic over Aes, added more KATs, and tried to match the repo's style.

### DIFF
--- a/ocb3/Cargo.toml
+++ b/ocb3/Cargo.toml
@@ -1,0 +1,42 @@
+[package]
+name = "ocb3"
+version = "0.1.0"
+description = """
+Pure Rust implementation of the AES-OCB3
+Authenticated Encryption with Associated Data (AEAD) Cipher
+"""
+authors = ["RustCrypto Developers"]
+edition = "2021"
+license = "Apache-2.0 OR MIT"
+readme = "README.md"
+documentation = "https://docs.rs/ocb3"
+repository = "https://github.com/RustCrypto/AEADs"
+keywords = ["aead", "aes", "encryption", "ocb", "ocb3"]
+categories = ["cryptography", "no-std"]
+rust-version = "1.56"
+
+[dependencies]
+aead = { version = "0.5", default-features = false }
+aes = { version = "0.8", optional = true }
+cipher = "0.4"
+ctr = "0.9"
+subtle = { version = "2", default-features = false }
+zeroize = { version = "1", optional = true, default-features = false }
+
+[dev-dependencies]
+aead = { version = "0.5", features = ["dev"], default-features = false }
+hex-literal = "0.3"
+
+[features]
+default   = ["aes", "alloc", "getrandom"]
+std       = ["aead/std", "alloc"]
+alloc     = ["aead/alloc"]
+arrayvec  = ["aead/arrayvec"]
+getrandom = ["aead/getrandom", "rand_core"]
+heapless  = ["aead/heapless"]
+rand_core = ["aead/rand_core"]
+stream    = ["aead/stream"]
+
+[package.metadata.docs.rs]
+all-features = true
+rustdoc-args = ["--cfg", "docsrs"]
--- a/ocb3/Cargo.toml
+++ b/ocb3/Cargo.toml
@@ -2,7 +2,7 @@
 name = "ocb3"
 version = "0.1.0"
 description = """
-Pure Rust implementation of the AES-OCB3
+Pure Rust implementation of the OCB3
 Authenticated Encryption with Associated Data (AEAD) Cipher
 """
 authors = ["RustCrypto Developers"]
@@ -11,7 +11,7 @@ license = "Apache-2.0 OR MIT"
 readme = "README.md"
 documentation = "https://docs.rs/ocb3"
 repository = "https://github.com/RustCrypto/AEADs"
-keywords = ["aead", "aes", "encryption", "ocb", "ocb3"]
+keywords = ["aead", "encryption", "ocb", "ocb3"]
 categories = ["cryptography", "no-std"]
 rust-version = "1.56"
 
--- a/ocb3/README.md
+++ b/ocb3/README.md
@@ -35,7 +35,7 @@ dual licensed as above, without any additional terms or conditions.
 [//]: # (badges)
 
 [crate-image]: https://buildstats.info/crate/ocb3
-[crate-link]: https://crates.io/crates/chacha20poly1305
+[crate-link]: https://crates.io/crates/ocb3
 [docs-image]: https://docs.rs/ocb3/badge.svg
 [docs-link]: https://docs.rs/ocb3/
 [license-image]: https://img.shields.io/badge/license-Apache2.0/MIT-blue.svg
--- a/ocb3/README.md
+++ b/ocb3/README.md
@@ -1,0 +1,51 @@
+# RustCrypto: OCB3
+
+[![crate][crate-image]][crate-link]
+[![Docs][docs-image]][docs-link]
+![Apache2/MIT licensed][license-image]
+![Rust Version][rustc-image]
+[![Project Chat][chat-image]][chat-link]
+[![Build Status][build-image]][build-link]
+
+Pure Rust implementation of **OCB3** ([RFC 7253][rfc7253])[Authenticated Encryption with Associated Data (AEAD)][aead] cipher.
+
+[Documentation][docs-link]
+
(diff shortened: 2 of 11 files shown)


ocb3/src/lib.rs
```diff
@@ -0,0 +1,604 @@
...
+mod util;
+use crate::util::{double, inplace_xor, ntz};
+use subtle::ConstantTimeEq;
+
+/// Number of L values to be precomputed. Precomputing m values, allows
+/// processing inputs of length up to 2^m blocks (2^m * 16 bytes) without
+/// needing to calculate L values at runtime.
+///
+/// By setting this to 32, we can process inputs of length up to 1 terabyte.
+const L_TABLE_SIZE: usize = 32;
```
sgmenda, line 32: Should we support inputs longer than one terabyte?

ocb3/README.md
```diff
@@ -0,0 +1,51 @@
...
+### Contribution
+
+Unless you explicitly state otherwise, any contribution intentionally submitted
+for inclusion in the work by you, as defined in the Apache-2.0 license, shall be
+dual licensed as above, without any additional terms or conditions.
+
+[//]: # (badges)
+
+[crate-image]: https://buildstats.info/crate/ocb3
+[crate-link]: https://crates.io/crates/chacha20poly1305
```
newpavlov, line 38: Wrong link.

ocb3/tests/kats.rs
```diff
@@ -0,0 +1,250 @@
...
+        plaintext: Vec<u8>,
+        ciphertext: Vec<u8>,
+    }
+    let kats = [
+            Kat {
+                nonce: hex!("BBAA99887766554433221100").to_vec(),
+                associated_data: hex!("").to_vec(),
+                plaintext: hex!("").to_vec(),
+                ciphertext: hex!("785407BFFFC8AD9EDCC5520AC9111EE6").to_vec(),
+            },
```
newpavlov, line 27: These tests are better implemented using the `new_test` macro. If you want, I can generate blobby file for these tests a bit later.

ocb3/Cargo.toml
```diff
@@ -0,0 +1,42 @@
+[package]
+name = "ocb3"
+version = "0.1.0"
+description = """
+Pure Rust implementation of the AES-OCB3
+Authenticated Encryption with Associated Data (AEAD) Cipher
```
newpavlov, line 6: I think we can remove the mention of AES and instead say that it's an implementation generic over block ciphers.

ocb3/src/util.rs
```diff
@@ -0,0 +1,76 @@
...
+        ]
+    }
+}
+
+/// Adapted from https://doc.rust-lang.org/std/primitive.slice.html#method.split_at_mut_unchecked
+///
+/// SAFETY: Assumes that `four_blocks` is exactly four blocks.
+#[inline]
+#[allow(unsafe_code)]
+pub(crate) unsafe fn split_into_four_blocks(four_blocks: &mut [u8]) -> [&mut Block; 4] {
```
newpavlov, line 65: I think we can make `split_into_two_blocks` and `split_into_four_blocks` safe functions. You use them in conjunction with `chunks_exact_mut`, so compiler can easily infer that the input arguments have the desired length, thus eliminating bound checking or asserts.

ocb3/src/lib.rs
```diff
@@ -0,0 +1,602 @@
...
+        #[cfg(target_feature = "avx512vaes")]
+        let split_into_blocks = crate::util::split_into_four_blocks;
+
+        let mut i = 1;
+
+        let mut Offset_i = [Block::default(); WIDTH];
+        Offset_i[Offset_i.len() - 1] = initial_offset(&self.cipher, nonce, TagSize::to_u32());
+        let mut Checksum_i = Block::default();
+        #[allow(unsafe_code)]
+        unsafe {
```
newpavlov, line 338: I don't think it makes much sense to add `#![deny(unsafe_code)]` and then `#[allow(unsafe_code)]` it for all `unsafe` blocks. Seems... kinda pointless, no? At worst, it can be seen as false advertisement: "look! our code does not have any unsafe! well, except here and hare".

ocb3/src/lib.rs
```diff
@@ -0,0 +1,602 @@
...
+    fn wide_encrypt(&self, nonce: &Nonce<NonceSize>, buffer: &mut [u8]) -> (usize, Block, Block) {
+        #[cfg(not(target_feature = "avx512vaes"))]
+        const WIDTH: usize = 2;
+        #[cfg(not(target_feature = "avx512vaes"))]
+        let split_into_blocks = crate::util::split_into_two_blocks;
+
+        #[cfg(target_feature = "avx512vaes")]
+        const WIDTH: usize = 4;
+        #[cfg(target_feature = "avx512vaes")]
+        let split_into_blocks = crate::util::split_into_four_blocks;
```
newpavlov, line 330: This code looks suspicious. Currently, the `aes` crate does not support VAES, and even if it did, you should use [the backend API](https://docs.rs/cipher/latest/cipher/trait.BlockEncrypt.html#tymethod.encrypt_with_backend) to take advantage of instruction-level parallelism supported by cipher implementation.

ocb3/src/lib.rs
```diff
@@ -0,0 +1,602 @@
+#![no_std]
+#![cfg_attr(docsrs, feature(doc_cfg))]
+#![doc = include_str!("../README.md")]
+#![doc(
+    html_logo_url = "https://raw.githubusercontent.com/RustCrypto/meta/master/logo.svg",
+    html_favicon_url = "https://raw.githubusercontent.com/RustCrypto/meta/master/logo.svg"
+)]
+#![deny(unsafe_code)]
+#![warn(missing_docs, rust_2018_idioms)]
+#![allow(non_snake_case)]
```
newpavlov, line 10: I would prefer for Rust code to follow the naming convention, even if it means using slightly different variable names from the spec.

ocb3/src/lib.rs
```diff
@@ -0,0 +1,602 @@
+#![no_std]
+#![cfg_attr(docsrs, feature(doc_cfg))]
+#![doc = include_str!("../README.md")]
```
newpavlov, line 3: If you include readme, it would be nice to add a simple example or two into it.

ocb3/src/lib.rs
```diff
@@ -0,0 +1,602 @@
...
+pub const C_MAX: usize = 1 << (L_TABLE_SIZE + 4);
+
+/// OCB3 nonce
+pub type Nonce<NonceSize> = GenericArray<u8, NonceSize>;
+
+/// OCB3 tag
+pub type Tag<TagSize> = GenericArray<u8, TagSize>;
+
+/// AES-OCB3 with a 128-bit key, 96-bit nonce, and 128-bit tag.
+pub type Aes128Ocb3 = AesOcb3<Aes128, U12>;
```
newpavlov, line 48: You don't feature gate these aliases properly. Personally, I would prefer for this crate to be cipher-agnostic, maybe later we could introduce `aes-ocb3` as a helper combination crate.

newpavlov (commented):
You should add a CI config for this crate.

Note: I haven't looked deep into the implementation yet.

sgmenda:
Thank you for the detailed comments.

I'd like to push back gently against decoupling OCB3 from AES. This implementation of OCB3 assumes a 128-bit blockcipher, which limits its utility with other blockciphers.

tarcieri:
@sgmenda you can express the block size requirement generically by bounding on the [`BlockSizeUser`](https://docs.rs/cipher/latest/cipher/trait.BlockSizeUser.html) trait, e.g.

```
Cipher: BlockCipher<BlockSize = U16>
```

One of the advantages of making the underlying cipher generic is that it becomes possible to replace the AES implementation with e.g. a hardware cryptographic accelerator. Embedded platforms which use such accelerators can benefit greatly from OCB because it cuts the number of cipher invocations in half versus e.g. the more commonly used CCM mode.

sgmenda:
@tarcieri Ah, TIL. I will make the change then.

dignifiedquire:
applied the missing CR in here: https://github.com/RustCrypto/AEADs/pull/587

tarcieri:
Merged in #587 

Thanks @sgmenda! Sorry it took so long!

Issue #496, opened 2022-12-23
AES-SIV-CMAC nonce sizes


davidv1992:
RFC 5297 specifies AES-CIV-CMAC as being able to accept any nonce size >= 1. However, the implementation here only supports 16 byte nonces. Would it be possible to support the other nonce sizes? I guess this would require a more general change to move closer to the aead interface specified in rfc 5116?

I am specifically asking for this as it would be required for me to support those different nonce sizes if I want to build a RFC-8915 compliant NTS server.

tarcieri:
It could potentially use a similar technique to the `aes-gcm` crate, which is generic around nonce sizes:

https://docs.rs/aes-gcm/latest/aes_gcm/struct.AesGcm.html

tarcieri:
Re: RFC5116, it was the original design impetus for the `aead` crate

davidv1992:
Hmm, that design is distinctly non-optimal for my usecase, as nonce length is really only known at runtime and is technically allowed to vary between 1 and infinity, and as such this would require a lot of instantiations of the types.

tarcieri:
If you really need a nonce whose length varies at runtime, I’d suggest building on top of the core `Siv` type:

https://docs.rs/aes-siv/latest/aes_siv/siv/struct.Siv.html

See the `SivAead` type for how the AEAD construction is built on top of it.

You can open an issue on the upstream traits repo to discuss traits for runtime-variable nonce lengths, though it’s obscure enough of a usecase and effectively an antipattern to the point that I’m not sure the additional complexity is worth it.

tarcieri:
Whoops, should probably leave this open to track potential support for type-level nonce size control

PR #555, opened 2023-10-10
feat(aes-siv): generic nonce sizes for SivAead

Direktor799:
add compile time NonceSize, #496

### DIFF
--- a/aes-siv/src/lib.rs
+++ b/aes-siv/src/lib.rs
@@ -91,12 +91,12 @@ pub use aead::{self, AeadCore, AeadInPlace, Error, Key, KeyInit, KeySizeUser};
 
 use crate::siv::Siv;
 use aead::{
-    consts::{U0, U16, U32, U64},
+    consts::{U0, U1, U16, U32, U64},
     generic_array::GenericArray,
     Buffer,
 };
 use aes::{Aes128, Aes256};
-use cipher::{ArrayLength, BlockCipher, BlockEncryptMut};
+use cipher::{typenum::IsGreaterOrEqual, ArrayLength, BlockCipher, BlockEncryptMut};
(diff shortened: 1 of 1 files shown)


aes-siv/src/lib.rs
```diff
@@ -105,20 +105,21 @@ use digest::{FixedOutputReset, Mac};
...
 /// commonly used Authenticated Encryption with Associated Data (AEAD) API,
 /// which accepts a key, nonce, and associated data when encrypting/decrypting.
-pub struct SivAead<C, M>
+pub struct SivAead<C, M, NonceSize = U16>
 where
     Self: KeySizeUser,
     C: BlockCipher<BlockSize = U16> + BlockEncryptMut + KeyInit + KeySizeUser,
     M: Mac<OutputSize = U16> + FixedOutputReset + KeyInit,
     <C as KeySizeUser>::KeySize: Add,
+    NonceSize: ArrayLength<u8>,
```
tarcieri, line 122: From [RFC5297 § 6.1](https://www.rfc-editor.org/rfc/rfc5297.html#section-6.1):

```
 N_MIN  is 1 octet.
```

So this should be reflected in the bounds, and probably documented somewhere:

```suggestion
    NonceSize: ArrayLength<u8> + IsGreaterOrEqual<U1>
```

tarcieri: Also:

> And the SivAead<C, M, NonceSize = U16> is not a specific AEAD algorithm mentioned in the RFC5297

The intent of `SivAead` is to implement the RFC5297 API, although the documentation to that effect may not be great.

Direktor799:
> This looks good now. I think it's backwards compatible?

yes, i think so

PR #558, opened 2023-10-25
gate test on feature it needs

alexanderkjall:
ref: https://salsa.debian.org/rust-team/debcargo-conf/-/blob/master/src/eax/debian/patches/gate-tests.patch

### DIFF
--- a/eax/tests/aes128eax.rs
+++ b/eax/tests/aes128eax.rs
@@ -4,4 +4,5 @@
 use aes::Aes128;
 use eax::Eax;
 
+#[cfg(feature = "std")]
 aead::new_test!(aes128eax, "aes128eax", Eax<Aes128>);


eax/tests/aes128eax.rs
```diff
@@ -4,4 +4,5 @@
 use aes::Aes128;
 use eax::Eax;
 
+#[cfg(feature = "std")]
```
tarcieri, line 7: I think this can just be...

```suggestion
#[cfg(feature = "alloc")]
```

tarcieri (approved):

eax/tests/aes128eax.rs
alexanderkjall: Thanks for the feedback, fixed

PR #562, opened 2023-10-31
add a test run with --no-default-features to eax

alexanderkjall:
to ensure that no error creep in by mistake


tarcieri:
Looks like that test is missing a directive to run itself when it changes. However, I can confirm it works locally, so I'm going to go ahead and merge.

PR #582, opened 2024-03-02
Support `no_unroll` for Ascon

korken89:
This helps when used on resource constrained systems, e.g. MCUs, to keep the binary size small.


newpavlov:
Closing similarly to https://github.com/RustCrypto/hashes/pull/567

PR #583, opened 2024-03-04
Bump `cipher` to `0.5.0-pre.4`

baloo:
Depends:
 - https://github.com/RustCrypto/traits/pull/1531
 - https://github.com/RustCrypto/universal-hashes/pull/196
 - https://github.com/RustCrypto/MACs/pull/158
 - https://github.com/RustCrypto/stream-ciphers/pull/344
 - https://github.com/RustCrypto/hybrid-array/pull/60

PR #587, opened 2024-03-24
ocb3: import from `offset-codebook-mode` crate

dignifiedquire:
I have applied the outstanding issues from #550 in here, hoping to get this merged and published.


tarcieri:
@dignifiedquire can you bump MSRV to 1.60?

wiktor-k:
Thanks for working on this @dignifiedquire !

OCB is "mandatory to implement" in the [new iteration of OpenPGP](https://www.ietf.org/archive/id/draft-ietf-openpgp-crypto-refresh-13.html#section-9.6) so this will definitely come in handy!

dignifiedquire:
> @dignifiedquire can you bump MSRV to 1.60?

just for the ocb crate for all crates?

tarcieri:
Just for ocb3 for now to get the tests passing

dignifiedquire:
ocb3 CI checks are now all passing

Cargo.lock
```diff
@@ -580,9 +593,9 @@ checksum = "9c8d87e72b64a3b4db28d11ce29237c246188f4f51057d65a7eab63b7987e423"
 
 [[package]]
 name = "zeroize"
-version = "1.6.0"
+version = "1.7.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2a0956f1ba7c7909bfb66c2e9e4124ab6f6482560f6628b5aaeba39207c9aad9"
+checksum = "525b4ec142c6b68a2d10f01f7bbf6755599ca3f81ea53b8431b7dd348f5fdb2d"
```
tarcieri, line 598: Can you revert this for now? It's breaking the build of the other crates.

We plan on bumping MSRV in #583 so all we're looking for now is a green build.

dignifiedquire:
now fully green :)

ocb3/Cargo.toml
```diff
@@ -0,0 +1,42 @@
...
+aead = { version = "0.5", default-features = false }
+cipher = "0.4"
+ctr = "0.9"
+subtle = { version = "2", default-features = false }
+zeroize = { version = "1", optional = true, default-features = false }
+
+[dev-dependencies]
+aead = { version = "0.5", features = ["dev"], default-features = false }
+aes = { version = "0.8", default-features = false }
+hex-literal = "0.3"
```
tarcieri, line 28: With the MSRV bump this can be:

```suggestion
hex-literal = "0.4"
```

ocb3/src/util.rs
```diff
@@ -0,0 +1,44 @@
...
+pub(crate) fn double(block: &Block) -> Block {
+    let mut v = u128::from_be_bytes((*block).into());
+    let v_hi = v >> 127;
+
+    // If v_hi = 0, return (v << 1)
+    // If v_hi = 1, return (v << 1) xor (0b0...010000111)
+    v <<= 1;
+    v ^= v_hi ^ (v_hi << 1) ^ (v_hi << 2) ^ (v_hi << 7);
+    v.to_be_bytes().into()
+}
```
tarcieri, line 30: I believe this polynomial is impl'd in the `dbl` crate, although I don't think we need to block a merge on that.

The reason `polyval` can't use that is because it's the reversed polynomial and little endian, whereas `dbl` is the big endian version.

ocb3/src/lib.rs
```diff
@@ -0,0 +1,588 @@
...
+///
+/// WARNING: Unless absolutely necessary, prefer the aliases Aes128Ocb3 and
+/// Aes256Ocb3.
+#[derive(Clone)]
+pub struct AesOcb3<Aes, NonceSize = U12, TagSize = U16>
+where
+    NonceSize: self::NonceSize,
+    TagSize: self::TagSize,
+{
+    cipher: Aes,
```
tarcieri, line 93: This should probably just be `Ocb3` declared generic around a cipher `C`. I can fix that up after a merge.

tarcieri: Also note `AesOcb3` can be achieved as a type alias so this all works out-of-the-box still.

PR #589, opened 2024-03-25
feat(ocb3): implement flexible nonce size

dignifiedquire:
Tested with nonce size `15` in rpgp. Unfortunately I don't have any additional test vectors.


ocb3/src/lib.rs
```diff
@@ -77,7 +77,21 @@ mod private {
...
+    impl SealedNonceSize for consts::U6 {}
+    impl SealedNonceSize for consts::U7 {}
+    impl SealedNonceSize for consts::U8 {}
+    impl SealedNonceSize for consts::U9 {}
+    impl SealedNonceSize for consts::U10 {}
+    impl SealedNonceSize for consts::U11 {}
     impl SealedNonceSize for consts::U12 {}
+    impl SealedNonceSize for consts::U13 {}
+    impl SealedNonceSize for consts::U14 {}
+    impl SealedNonceSize for consts::U15 {}
```
tarcieri, line 94: Hmm, should we change this to an [`LeEq`](https://docs.rs/typenum/latest/typenum/operator_aliases/type.LeEq.html) constraint instead?

tarcieri: The same probably goes for `TagSize`

dignifiedquire: yes that seems better

dignifiedquire: okay, I have no idea how to use this though to make a blanket impl 😅

tarcieri: I opened #590 to use `IsLessOrEqual` for both `NonceSize` and `TagSize`

dignifiedquire: rebased accordingly

tarcieri (approved):

PR #590, opened 2024-03-25
ocb3: use `IsLessOrEqual` for `NonceSize`/`TagSize`

tarcieri:
Changes the bounds from using a sealed trait to the `IsLessOrEqual` operator, which both better reflects the constraints as well as makes it easier to eventually translate to const generics.


dignifiedquire (approved):
nice, thanks

PR #593, opened 2024-03-27
ocb3: restrict minimum `NonceSize` to `U6`

### LINKED ISSUE #592: OCB3: restrict short nonces
This paper presents an attack on OCB3 with short nonces:

https://eprint.iacr.org/2023/326.pdf

>  This note reports a small flaw in the security proof of OCB3 that
> may cause a loss of security in practice, even if OCB3 is correctly implemented in a trustworthy
> and nonce-respecting module. The flaw is present when OCB3 is used with short nonces. It has
> security implications that are worse than nonce-repetition as confidentiality and authenticity are
> lost until the key is changed. The flaw is due to an implicit condition in the security proof and to
> the way OCB3 processes nonce.

It makes the following recommendation:

> In the case of OCB3, it is easy to fix the algorithm’s specification in order to avoid the
> weakness and abide to the full assumptions of the security proof. If the description is unchanged,
> the requirement N ≥ 6 must become an absolute requirement.

We should update the bounds on the nonce size accordingly.

tarcieri:
Based on the recommendations of the following paper, which describes an attack against OCB3 mode:

https://eprint.iacr.org/2023/326.pdf

> In the case of OCB3, it is easy to fix the algorithm’s specification
> in order to avoid the weakness and abide to the full assumptions of
> the security proof. If the description is unchanged, the requirement
> N ≥ 6 must become an absolute requirement.

Furthermore, this restricts the minimum tag size to 1-byte, up from the former 0-bytes. This is a questionable choice of minimum but reflects the wording in the RFC:

> The TAGLEN parameter specifies the length of authentication tag used
> by OCB and may be any value up to 128

Closes #592

### DIFF
--- a/ocb3/src/lib.rs
+++ b/ocb3/src/lib.rs
@@ -10,7 +10,7 @@
 
 /// Constants used, reexported for convenience.
 pub mod consts {
-    pub use cipher::consts::{U0, U12, U15, U16};
+    pub use cipher::consts::{U0, U12, U15, U16, U6};
 }
 
 mod util;
@@ -20,9 +20,12 @@ pub use aead::{
 };
 
 use crate::util::{double, inplace_xor, ntz, Block};
-use aead::generic_array::{typenum::IsLessOrEqual, ArrayLength};
+use aead::generic_array::{
+    typenum::{IsGreater, IsGreaterOrEqual, IsLessOrEqual},
+    ArrayLength,
+};
 use cipher::{
-    consts::{U0, U12, U15, U16},
+    consts::{U0, U12, U15, U16, U6},
     BlockDecrypt, BlockEncrypt, BlockSizeUser,
 };
 use core::marker::PhantomData;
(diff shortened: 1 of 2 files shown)


ocb3/src/lib.rs
```diff
@@ -79,17 +87,17 @@ type Sum = GenericArray<u8, SumSize>;
 impl<Cipher, NonceSize, TagSize> KeySizeUser for Ocb3<Cipher, NonceSize, TagSize>
 where
     Cipher: KeySizeUser,
-    TagSize: ArrayLength<u8> + IsLessOrEqual<U16>,
-    NonceSize: ArrayLength<u8> + IsLessOrEqual<U15>,
+    TagSize: ArrayLength<u8> + IsGreater<U0> + IsLessOrEqual<U16>,
+    NonceSize: ArrayLength<u8> + IsGreaterOrEqual<U6> + IsLessOrEqual<U15>,
```
newpavlov, line 91: I think you also need to add bounds like `GrEq<NonceSize, U6>: NonZero,`. Plus it's worth to add compilation failure tests for unsupported sizes.

newpavlov: It may be worth to add sealed `TagSizes`/`NonceSizes` traits to simplify public bounds.

newpavlov: I can prepare a quick PR to amend this.

tarcieri: We had those before although removed them in #590, though the requirements have gotten more complex since then. Still, it's a bounded range, and if we can express it properly with `typenum` it will be easier to eventually move to const generics.

PR #604, opened 2024-05-18
Implement belt dwp

makavity:
Implemented Belt-DWP, republic of Belarus AEAD algorithm.


makavity:
Hey, @newpavlov!
Could you check it, please? :)

makavity:
Looks like benches is broken in master

belt-dwp/Cargo.toml
```diff
@@ -0,0 +1,39 @@
...
+aead = { version = "0.6.0-rc.0", default-features = false }
+zeroize = { version = "1.7", default-features = false, optional = true }
+universal-hash = { version = "0.6.0-rc.0" }
+opaque-debug = { version = "0.3" }
+
+belt-block = { version = "0.2.0-pre.2" }
+belt-ctr = { version = "0.2.0-pre" }
+
+[dev-dependencies]
+hex-literal = "0.4"
```
newpavlov, line 24: Use `hex-literal` v1

belt-dwp/Cargo.toml
```diff
@@ -0,0 +1,39 @@
...
+bytes = ["aead/bytes"]
+os_rng = ["aead/os_rng", "rand_core"]
+heapless = ["aead/heapless"]
+rand_core = ["aead/rand_core"]
+reduced-round = []
+zeroize = ["dep:zeroize", "belt-ctr/zeroize"]
+
+[package.metadata.docs.rs]
+all-features = true
+rustdoc-args = ["--cfg", "docsrs"]
```
newpavlov, line 39: We don't need to pass `--cfg docsrs` anymore, since docs.rs passes this flag itself by default (we should remove it from other crates as well). Also, keep an empty last line.

belt-dwp/README.md
```diff
@@ -0,0 +1,59 @@
...
+[![Docs][docs-image]][docs-link]
+![Apache2/MIT licensed][license-image]
+![Rust Version][rustc-image]
+[![Project Chat][chat-image]][chat-link]
+[![Build Status][build-image]][build-link]
+
+Pure Rust implementation of **BeltDwp** ([STB 34.101.31-2020][1]): an
+[Authenticated Encryption with Associated Data (AEAD)][2].
+
+[Documentation][docs-link]
```
newpavlov, line 13: Since you use readme as crate docs, this doc link is not needed, since it looks strange in generated crate docs.

belt-dwp/tests/belt.rs
```diff
@@ -0,0 +1,47 @@
+use aead::AeadInPlaceDetached;
+use belt_dwp::{BeltDwp, KeyInit};
+use hex_literal::hex;
+
+/// Test from Appendix A, tables 19-20 of [STB 34.101.31-2020](https://apmi.bsu.by/assets/files/std/belt-spec372.pdf)
```
newpavlov, line 5: Since these doc strings do not get rendeed, you can write it like this:
```rust
/// Test from Appendix A, tables 19-20 of STB 34.101.31-2020:
/// https://apmi.bsu.by/assets/files/std/belt-spec372.pdf
```

belt-dwp/src/utils.rs
```diff
@@ -0,0 +1,29 @@
+/// Helper function for transforming BelT keys and blocks from a byte array
+/// to an array of `u32`s.
+///
+/// # Panics
+/// If length of `src` is not equal to `4 * N`.
+// #[inline(always)]
```
newpavlov, line 6: Uncomment or remove completely?

belt-dwp/src/utils.rs
```diff
@@ -0,0 +1,29 @@
...
+        .for_each(|(dst, src)| *dst = u32::from_le_bytes(src.try_into().unwrap()));
+    res
+}
+
+/// Helper function for transforming BelT keys and blocks from a array of `u32`s
+/// to a byte array.
+///
+/// # Panics
+/// If length of `src` is not equal to `4 * N`.
+// #[inline(always)]
```
newpavlov, line 21: Same.

belt-dwp/src/utils.rs
```diff
@@ -0,0 +1,29 @@
...
+pub(crate) fn to_u32<const N: usize>(src: &[u8]) -> [u32; N] {
+    assert_eq!(src.len(), 4 * N);
+    let mut res = [0u32; N];
+    res.iter_mut()
+        .zip(src.chunks_exact(4))
+        .for_each(|(dst, src)| *dst = u32::from_le_bytes(src.try_into().unwrap()));
```
newpavlov, line 12: `array::from_fn` can help to write it in a more concise manner:
```rust
core::array::from_fn(|i| {
    let bytes = src[4 * i...][..4].try_into().unwrap();
    u32::from_le_bytes(bytes);
})
```

belt-dwp/src/lib.rs
```diff
@@ -0,0 +1,276 @@
...
+
+/// Tag type for [`BeltDwp`]
+pub type Tag = aead::Tag<BeltDwp>;
+
+mod gf;
+mod ghash;
+mod utils;
+
+/// T from the STB 34.101.31-2020
+const T: u128 = 0xE45D4A588E006D363BF5080AC8BA94B1;
```
newpavlov, line 103: Use underscore for better readability, i.e. `0xE45D_4A58_8E00_6D36_3BF5_080A_C8BA_94B1`.

belt-dwp/src/lib.rs
```diff
@@ -0,0 +1,276 @@
...
+
+/// Belt-DWP authenticated encryption with associated data (AEAD) cipher, defined in
+/// STB 34.101.31-2020
+pub struct BeltDwp {
+    key: Key<BeltBlock>,
```
newpavlov, line 108: It's better to keep initialized `BeltBlock` in `BeltDwp` and pre-computed `r` and `s` constants instead of the raw key. Note that you should be able to create and use an instance of `BeltCtr<&BeltBlock>`.

belt-dwp/src/lib.rs
```diff
@@ -0,0 +1,276 @@
...
+        self.ghash.update_padded(associated_data);
+
+        // 4. For 𝑖 = 1, 2, . . . , 𝑛 do:
+        //  4.1 𝑠 ← 𝑠 ⊞ ⟨1⟩_128
+        //  4.2 𝑌𝑖 ← 𝑋𝑖 ⊕ Lo(belt-block(𝑠, 𝐾), |𝑋𝑖|)
+        //  4.3 𝑡 ← 𝑡 ⊕ (𝑌𝑖 ‖ 0^{128−|𝑌𝑖|})
+        //  4.4 𝑡 ← 𝑡 * 𝑟.
+        buffer.chunks_mut(16).for_each(|block| {
+            self.enc_cipher.apply_keystream(block);
+            self.ghash.update_padded(block);
```
newpavlov, line 183: You should use `BeltCtrCore::process_with_backend` here.

belt-dwp/src/lib.rs
```diff
@@ -0,0 +1,276 @@
...
+        // 5. 𝑡 ← 𝑡 ⊕ (⟨|𝐼|⟩_64 ‖ ⟨|𝑋|⟩_64)
+        self.ghash.xor_s(&sizes_block);
+
+        let mut tag_exact = self.finish_tag();
+
+        // 6. 𝑡 ← belt-block(𝑡 * 𝑟, 𝐾).
+        self.mac_cipher.encrypt_block(&mut tag_exact);
+
+        // 7. If 𝑇 != Lo(𝑡, 64), return ⊥
+        if tag_exact[..8] != tag[..] {
```
newpavlov, line 225: You should use `subtle` here to perform equality check in a const-time fashion.

belt-dwp/src/lib.rs
```diff
@@ -0,0 +1,276 @@
...
+            return Err(aead::Error);
+        }
+
+        // 8. For 𝑖 = 1,2,...,𝑛 do:
+        // 8.1. 𝑠 ← 𝑠 ⊞ ⟨1⟩128;
+        // 8.2. 𝑋𝑖 ← 𝑌𝑖 ⊕ Lo(belt-block(𝑠, 𝐾), |𝑌𝑖|)
+        buffer.chunks_mut(16).for_each(|block| {
+            self.enc_cipher.apply_keystream(block);
+        });
```
newpavlov, line 234: You don't need to chunk data here. You can simply call `apply_keystream` on `buffer`. Also note that one-pass processing could be more efficient (i.e. you would update the `ghash` state and decrypt on each block), but in the case of tag mismatch you would need to either re-encrypt or zeroize output buffer (see #660).

belt-dwp/src/ghash.rs
```diff
@@ -0,0 +1,77 @@
...
+pub type Key = universal_hash::Key<GHash>;
+
+/// GHASH blocks (16-bytes)
+pub type Block = universal_hash::Block<GHash>;
+
+/// GHASH tags (16-bytes)
+pub type Tag = universal_hash::Block<GHash>;
+
+#[derive(Clone)]
+pub struct GHash {
```
newpavlov, line 17: `GHash` is not exposed as part of public API, so you could use inherent methods here without relying on the `universal-hash` dependency.

newpavlov (commented):
Sorry for the late review!

belt-dwp/Cargo.toml
makavity: Done

belt-dwp/src/lib.rs
makavity: I didn't really understand how to do this, because s and r are created at the stage with the transfer of Nonce. That's why I did it this way.

belt-dwp/src/lib.rs
makavity: I also don't really understand how to do this better? I should create a separate structure in which to place my blocks and implement StreamCipherBackend?

belt-dwp/src/lib.rs
makavity: Thank you! Done.
Looks like i don't need to re-encrypt, because encryption is only applied after tag calculating?
It is not so optimized, but looks fine, as I suggest.

belt-dwp/src/ghash.rs
makavity: Done, thank you!

belt-dwp/Cargo.toml
newpavlov: You should keep `all-features = true`.

belt-dwp/src/lib.rs
newpavlov: Ah, true. Then just use initialized `BeltBlock` instead of the raw key.

belt-dwp/src/ghash.rs
newpavlov: You only changed the type aliases. I meant that you don't need to implement the `universal-hash` traits and you can remove it from the dependency list.

belt-dwp/src/lib.rs
newpavlov: Yes, with two-pass decryption you don't need to re-encrypt. We can keep the current implementation and leave one-pass implementation for a potential later PR.

belt-dwp/src/lib.rs
```diff
@@ -0,0 +1,274 @@
...
+        // 5. 𝑡 ← 𝑡 ⊕ (⟨|𝐼|⟩_64 ‖ ⟨|𝑋|⟩_64)
+        self.ghash.xor_s(&sizes_block);
+
+        // 6. 𝑡 ← belt-block(𝑡 * 𝑟, 𝐾).
+        let mut tag = self.finish_tag();
+
+        self.mac_cipher.encrypt_block(&mut tag);
+
+        // Unwrap is safe because the tag is always 8 bytes
+        Ok(Tag::try_from(&tag[..8]).unwrap())
```
newpavlov, line 195: Instead of the comment it's better to use `expect`.

belt-dwp/src/lib.rs
```diff
@@ -0,0 +1,274 @@
...
+        }
+    }
+
+    fn encrypt_in_place_detached(
+        &mut self,
+        associated_data: &[u8],
+        buffer: &mut [u8],
+    ) -> aead::Result<Tag> {
+        let sizes_block =
+            get_sizes_block(associated_data.len() as u64 * 8, buffer.len() as u64 * 8);
```
newpavlov, line 169: Introduce separate variables to make it a bit more readable?
```rust
let aad_bit_len = associated_data.len() as u64 * 8;
let msg_bit_len = buffer.len() as u64 * 8;
```

belt-dwp/src/lib.rs
newpavlov: You should use `BeltCtrCore::inner_iv_init(&self.cipher, &nonce)`.

belt-dwp/src/lib.rs
newpavlov: Not quite, you need to implement the `StreamClosure` trait, which would accept `StreamCipherBackend` (effectively, you need to manually implement closure `|stream_backend| { ... }`).

I guess we can keep the current code as-is for now and migrate to `BeltCtrCore` later in a separate PR.

belt-dwp/Cargo.toml
```diff
@@ -0,0 +1,36 @@
...
+zeroize = { version = "1.7", default-features = false, optional = true }
+universal-hash = { version = "0.6.0-rc.0" }
+opaque-debug = { version = "0.3" }
+subtle = { version = "2", default-features = false }
+
+belt-block = { version = "0.2.0-pre.2" }
+belt-ctr = { version = "0.2.0-pre" }
+
+[dev-dependencies]
+hex-literal = "1.0.0"
```
newpavlov, line 25: Use `hex-literal = "1"`.

belt-dwp/src/ghash.rs
newpavlov: On the second thought, we can leave it as-is with the goal of future migration to the `ghash` crate.

belt-dwp/src/lib.rs
makavity: Looks like for that, I need to copy BeltBlock on each operation. Is it fine?

belt-dwp/src/lib.rs
makavity: Sure. Fixed.

belt-dwp/src/lib.rs
makavity: Also fixed. Thanks

belt-dwp/Cargo.toml
makavity: Fixed

belt-dwp/src/lib.rs
newpavlov: No, you can create a `BeltCtr` instance using reference to the block cipher (note the `&self.cipher` part). In other words, you will use `BeltCtr<&BeltBlock>`, not `BeltCtr<BeltBlock>`.

makavity: Nvm. Just removed Cipher. Don’t thinks it is needed.

belt-dwp/src/lib.rs
```diff
@@ -0,0 +1,249 @@
...
+mod gf;
+mod ghash;
+
+/// T from the STB 34.101.31-2020
+const T: u128 = 0xE45D_4A58_8E00_6D36_3BF5_080A_C8BA_94B1;
+
+/// Belt-DWP authenticated encryption with associated data (AEAD) cipher, defined in
+/// STB 34.101.31-2020
+pub struct BeltDwp {
+    backend: BeltBlock,
```
newpavlov, line 105: It may be worth to make this implementation generic over 128-bit block ciphers to allow plugging a different implementation. You can keep `BeltBlock` as the default type similarly to what was done in `belt-ctr`.

belt-dwp/src/lib.rs
```diff
@@ -15,9 +15,9 @@
 #![cfg_attr(not(all(feature = "os_rng", feature = "heapless")), doc = "```ignore")]
 //! # fn main() -> Result<(), Box<dyn std::error::Error>> {
 //! use belt_dwp::{
-//!     aead::{Aead, AeadCore, KeyInit},
-//!     BeltDwp, Nonce
+//!     aead::{Aead, AeadCore, KeyInit}, Nonce, BeltBlock
 //! };
+//! type BeltDwp = belt_dwp::BeltDwp::<BeltBlock>;
```
newpavlov, line 20: In the examples you can just use `BeltDwp` without defining this explicit alias.

belt-dwp/tests/belt.rs
```diff
@@ -1,7 +1,10 @@
 use aead::AeadInPlaceDetached;
-use belt_dwp::{BeltDwp, KeyInit};
+use belt_block::BeltBlock;
+use belt_dwp::KeyInit;
 use hex_literal::hex;
 
+type BeltDwp = belt_dwp::BeltDwp<BeltBlock>;
```
newpavlov, line 6: Same here, just use `BeltDwp` with the default type parameter.

makavity:
```rust
   Compiling belt-dwp v0.1.0 (/Users/makavity/dev/crypto/AEADs/belt-dwp)
error[E0283]: type annotations needed for `BeltDwp<_>`
   --> belt-dwp/tests/belt.rs:39:13
    |
39  |         let belt_dwp = BeltDwp::new_from_slice(&vec.k).unwrap();
    |             ^^^^^^^^   ------- type must be known at this point
    |
    = note: cannot satisfy `_: cipher::block::BlockCipherEncrypt`
    = help: the following types implement trait `cipher::block::BlockCipherEncrypt`:
              &Alg
              BeltBlock
              cipher::tweak::zero::ZeroTweak<C>
note: required by a bound in `BeltDwp`
   --> /Users/makavity/dev/crypto/AEADs/belt-dwp/src/lib.rs:106:8
    |
104 | pub struct BeltDwp<C = BeltBlock>
    |            ------- required by a bound in this struct
105 | where
106 |     C: BlockCipherEncrypt + BlockSizeUser<BlockSize = U16>,
    |        ^^^^^^^^^^^^^^^^^^ required by this bound in `BeltDwp`
help: consider giving `belt_dwp` an explicit type, where the type for type parameter `C` is specified
    |
39  |         let belt_dwp: BeltDwp<C> = BeltDwp::new_from_slice(&vec.k).unwrap();
    |                     ++++++++++++

For more information about this error, try `rustc --explain E0283`.
error: could not compile `belt-dwp` (test "belt") due to 1 previous error
```

Should I use 
```rust
let dwp: BeltDwp = BeltDwp::new_from_slice(&vec.k).unwrap();
```
?

newpavlov: Huh... weird. I guess one way to work around this issue is to write this:
```rust
pub struct Dwp<C> {
    cipher: C,
}

pub type BeltDwp = Dwp<BeltBlock>;
```

Though I am interested in learning why the default type parameter does not work here as expected.

makavity: The same problem also in belt-car crate

newpavlov: Yeah, it's probably worth to define a generic `Ctr` type and `BeltCtr = Ctr<BeltBlock>` type alias on top of it.

makavity: Okay, thank you.
Fixed generic default by type alias, as you suggest.

newpavlov (approved):

newpavlov:
Thank you!

PR #606, opened 2024-05-31
ocb3: use `dbl` crate

tarcieri:
Replaces the vendored implementation of GF(2^128) doubling (i.e. multiply-by-x) with the `dbl` crate.

Also moves the remaining functions in the `util` module into lib.rs.


newpavlov:
IIRC you argued that we should deprecate the `dbl` crate and instead prefer the (relatively simple) vendored implementations. Have you changed your mind about this?

tarcieri:
Hmm, tough call. On the one hand it is a pattern that appears in a lot of places, although right now the polynomial is hardcoded. On the other we don't provide optimized implementations so the implementation is trivial.

tarcieri:
I think we should go ahead and merge this, and if we decide to retire `dbl`, vendor the exact implementation in there into the respective crates. Perhaps that needs a tracking issue?

tarcieri:
I made a tracking issue about potentially retiring `dbl`: https://github.com/RustCrypto/utils/issues/1077

In the meantime I think we should move forward with this, and rip it out if we decide to rip it out everywhere.

Issue #1, opened 2019-08-16
Missing algorithms


newpavlov:
TODO

tarcieri:
@newpavlov I have a locally working `chacha20poly1305` crate I can push up, but I don't have permission.

It should be fairly trivial to implement both AES-GCM and AES-GCM-SIV once I have my implementation of POLYVAL working:

https://github.com/RustCrypto/MACs/pull/13

newpavlov:
Ah, I forgot to add this repository to the team.

warner:
I'll throw in a request for XSalsa20Poly1305. I'm looking [to replace](https://github.com/warner/magic-wormhole.rs/issues/58) magic-wormhole's libsodium dependency with something smaller, but I need to retain interoperability with the default libsodium `secretbox` implementation, which uses XSalsa20 and not XChaCha20.

tarcieri:
AES-GCM and XSalsa20Poly1305 are now done :shipit:

I also have a WIP PR to merge the AES-SIV implementation from Miscreant

zer0x64:
It would be very cool to add support for CAESAR competition winners:
https://competitions.cr.yp.to/caesar-submissions.html

Even though they are not widely used, they are considered the "best option if available" and they would give an edge to Rust, especially considering how easy cross-platform Rust is.

According to the page, ACORN and COLM are considered "second-choice" so I believe they should also come second in an order of priority. So the ciphers to implement first would be:
- Ascon
- AEGIS-128(/256?) and/or OCB
- Deoxys-II

zer0x64:
Another suggestion would be XChaCha20-Poly1305. 

The reason is that, if there is a lot of encryption/decryption with the same key, with standard ChaCha20 might be vulnerable to a nonce collision. A single collision is enough to break the authenticity provided by Poly1305. 

The main difference between the two is that XChaCha20 uses 192 bits nonce instead of 64 bits nonce, which makes collisions completely impractical if properly generated. Since there is already a crate for ChaCha20 and XSalsa20, I guess it wouldn't be really hard to implement.

tarcieri:
@zer0x64 it's already implemented in the `chacha20poly1305` crate:

https://docs.rs/chacha20poly1305/latest/chacha20poly1305/struct.XChaCha20Poly1305.html

zer0x64:
Oh, didn't saw that! Thanks for clarifying!

elichai:
>     * [x]  AES-GCM
> 
[...]

PR #612, opened 2024-06-29
xaes: initial implementation

SergioBenitez:
This is an initial implementation of [XAES-256-GCM](https://github.com/C2SP/C2SP/blob/main/XAES-256-GCM.md) (re: https://github.com/RustCrypto/AEADs/issues/1) which passes the test vectors.

Would love a review, especially as it pertains to constant-time and zeroing (why isn't `zeroize` used to zero IVs?). I don't see an obvious constant-time byte-slice XOR in use elsewhere in Rust-Crypto, but please point to a canonical reference if possible. I also have not placed this behind any feature flags, yet. Finally, the primary structure `XaesGcm256` is not parameterized in any way. If it's desirable to parameterize it in a similar fashion to `AesGcm`, please let me know.


aes-gcm/src/xaes.rs
```diff
@@ -0,0 +1,127 @@
...
+type NonceSize = <<Aes256Gcm as AeadCore>::NonceSize as Mul<U2>>::Output;
+type TagSize = <Aes256Gcm as AeadCore>::TagSize;
+type CiphertextOverhead = <Aes256Gcm as AeadCore>::CiphertextOverhead;
+type Block = Array<u8, <Aes256 as BlockSizeUser>::BlockSize>;
+
+/// Maximum length of plaintext.
+pub const P_MAX: u64 = 1 << 36;
+
+/// Maximum length of associated data.
+// pub const A_MAX: u64 = 1 << 61;
```
pinkforest, line 25: probably forgot to delete ?

aes-gcm/src/lib.rs
```diff
@@ -108,11 +108,17 @@
...
 
+#[cfg(feature = "aes")]
+mod xaes;
+
 pub use aead::{self, AeadCore, AeadInPlace, Error, Key, KeyInit, KeySizeUser};
 
 #[cfg(feature = "aes")]
 pub use aes;
 
+#[cfg(feature = "aes")]
```
pinkforest, line 119: Any reason it is under feature = "aes" ? Any reason it can't be separate feature = "xaes" extending "aes" ?

aes-gcm/tests/xaes256gcm.rs
```diff
@@ -0,0 +1,35 @@
...
+
+#[macro_use]
+mod common;
+
+use aes_gcm::aead::{array::Array, Aead, AeadInPlace, KeyInit, Payload};
+use aes_gcm::XaesGcm256;
+use common::TestVector;
+use hex_literal::hex;
+
+/// C2SP XAES-256-GCM test vectors
```
pinkforest, line 13: sidebar - would love if C2SP test vectors would be embedded in crate a-like of [wycheproof-rs](https://crates.io/crates/wycheproof) so test vectors can be updated independently to all associated impl's and no messing with blobs / updating multiple places

aes-gcm/src/xaes.rs
```diff
@@ -0,0 +1,127 @@
...
+        let mut m2 = Block::default();
+        m2[..4].copy_from_slice(&[0, 2, b'X', 0]);
+        m2[4..].copy_from_slice(n1);
+
+        // Kₘ = AES-256ₖ(M1 ⊕ K1)
+        // Kₙ = AES-256ₖ(M2 ⊕ K1)
+        // Kₓ = Kₘ || Kₙ = AES-256ₖ(M1 ⊕ K1) || AES-256ₖ(M2 ⊕ K1)
+        let mut key: Key<Aes256Gcm> = Array::default();
+        let (km, kn) = key.split_ref_mut::<<KeySize as Div<U2>>::Output>();
+        for i in 0..km.len() {
```
pinkforest, line 116: Would be good to comment why conditional `km.len()` is okay to potentially leak on all conditionals ?

aes-gcm/src/xaes.rs
```diff
@@ -0,0 +1,127 @@
...
+    fn new(key: &Key<Aes256>) -> Self {
+        let aes = Aes256::new(key);
+
+        // L = AES-256ₖ(0¹²⁸)
+        let mut k1 = Block::default();
+        aes.encrypt_block(&mut k1);
+
+        // If MSB₁(L) = 0 then K1 = L << 1 Else K1 = (L << 1) ⊕ 0¹²⁰10000111
+        let mut msb = 0;
+        for i in (0..k1.len()).rev() {
```
pinkforest, line 52: Also probably good to document `k1.len()` conditional here

pinkforest:
gave you few nits - take em if you like :relaxed: 

if you feel like it and if worried about non-ct anywhere:
- https://github.com/Ledger-Donjon/cargo-checkct | https://www.ledger.com/blog-cargo-checkct-our-home-made-tool-guarding-against-timing-attacks-is-now-open-source
- https://crates.io/crates/dudect-bencher - needs a crafted dataset though [crypto-bigint](https://github.com/RustCrypto/crypto-bigint/tree/master/dudect) | [basic example](https://github.com/pinkforest/investigate-mir-secret-data-leak/tree/main/side_channels_dudect)

aes-gcm/src/xaes.rs
```diff
@@ -0,0 +1,127 @@
...
+    }
+}
+
+impl XaesGcm256 {
+    // Implements steps 3 - 5 of the spec.
+    fn derive_key(&self, n1: &Nonce<<NonceSize as Div<U2>>::Output>) -> Key<Aes256Gcm> {
+        // M1 = 0x00 || 0x01 || X || 0x00 || N[:12]
+        let mut m1 = Block::default();
+        m1[..4].copy_from_slice(&[0, 1, b'X', 0]);
+        m1[4..].copy_from_slice(n1);
```
pinkforest, line 104: conditional length on ? would be great to clarify maybe why perhaps

aes-gcm/tests/common/mod.rs
```diff
@@ -4,7 +4,7 @@
 #[derive(Debug)]
 pub struct TestVector<K: 'static> {
     pub key: &'static K,
-    pub nonce: &'static [u8; 12],
+    pub nonce: &'static [u8],
```
pinkforest, line 7: Is there a reason to make this non-fixed length ?

aes-gcm/src/xaes.rs
```diff
@@ -0,0 +1,127 @@
...
+
+        // L = AES-256ₖ(0¹²⁸)
+        let mut k1 = Block::default();
+        aes.encrypt_block(&mut k1);
+
+        // If MSB₁(L) = 0 then K1 = L << 1 Else K1 = (L << 1) ⊕ 0¹²⁰10000111
+        let mut msb = 0;
+        for i in (0..k1.len()).rev() {
+            let new_msb = k1[i] >> 7;
+            k1[i] = (k1[i] << 1) | msb;
```
pinkforest, line 54: Should this be in [subtle::Blackbox::new()](https://docs.rs/subtle/latest/subtle/struct.BlackBox.html) ref https://github.com/dalek-cryptography/curve25519-dalek/pull/662/files

aes-gcm/src/xaes.rs
```diff
@@ -0,0 +1,127 @@
...
+        // Kₘ = AES-256ₖ(M1 ⊕ K1)
+        // Kₙ = AES-256ₖ(M2 ⊕ K1)
+        // Kₓ = Kₘ || Kₙ = AES-256ₖ(M1 ⊕ K1) || AES-256ₖ(M2 ⊕ K1)
+        let mut key: Key<Aes256Gcm> = Array::default();
+        let (km, kn) = key.split_ref_mut::<<KeySize as Div<U2>>::Output>();
+        for i in 0..km.len() {
+            km[i] = m1[i] ^ self.k1[i];
+        }
+        for i in 0..kn.len() {
+            kn[i] = m2[i] ^ self.k1[i];
```
pinkforest, line 120: Also should these be in [subtle::BlackBox](https://docs.rs/subtle/latest/subtle/struct.BlackBox.html)

aes-gcm/tests/common/mod.rs
SergioBenitez: Now we have nonces of two different lengths, so this is needed to be able to reuse the structure.

aes-gcm/src/xaes.rs
SergioBenitez: I left the code here to inspire this comment. :)

According to the C2SP spec, the associated data max size for XAES-256-GCM is 2EiB = 2^61 bytes. However the max size for AD for AES-GCM is 2^64 and yet it's been restricted to 2^36 by this crate, so specifying the "true" AD would have no affect. I didn't explore why the AD is limited in this manner, however, so I left the code there in hopes that that would be clarified.

aes-gcm/src/lib.rs
SergioBenitez: Whatever is desired!

aes-gcm/src/xaes.rs
SergioBenitez: Not sure I follow. `km.len()` isn't a conditional. Perhaps I'm misunderstanding?

aes-gcm/src/xaes.rs
SergioBenitez: Same as above.

aes-gcm/src/xaes.rs
SergioBenitez: I don't understand the comment. What is `?`?

aes-gcm/src/xaes.rs
SergioBenitez: I compiled this with various flags, and at least of the nightly as of 3 days, the assembly generated is identical and doesn't leak any secrets via timing side-channels. That may change with future compilers, of course.

Where precisely are you suggesting `Blackbox` be used? Could you provide a GitHub suggestion of the change, perhaps?

SergioBenitez:
@newpavlov @tarcieri Any chance for a review here?

SergioBenitez:
Checking in. Any chance to push this forward?

tarcieri:
@SergioBenitez haven't had a whole lot of free time lately for code review but I still hope to review it soon

tarcieri:
Sorry for the belated review.

On #1 we had discussed an `xaes-gcm` crate, but I now see there is [already an `xaes-gcm` crate](https://crates.io/crates/xaes-gcm) which implements the [Derive-Key-AES-GCM](https://csrc.nist.gov/csrc/media/Presentations/2023/constructions-based-on-the-aes-round/images-media/sess-5-gueron-bcm-workshop-2023.pdf) construction, and I have not carefully looked at the differences between the protocols.

I am a bit wary including the construction in the `aes-gcm` crate itself, which otherwise implements NIST standard constructions.

However, I'd also note the construction in the spec is called [`XAES-256-GCM`](https://github.com/C2SP/C2SP/blob/main/XAES-256-GCM.md), so how about an `xaes-256-gcm` crate instead, which is currently available?

SergioBenitez:
Sure! Went ahead and published a `-pre` version to reserve the name. I also added you as an owner; feel free to use that as you wish. Looking forward to a review of the crate.

tarcieri (approved):
Sorry for the long delay in review! This looks good now.

tarcieri:
@SergioBenitez attempted to do a Cargo.lock merge but it seems it didn't work. Can you take a look?

tarcieri:
Merged in #642

PR #625, opened 2024-09-10
Implement trait AeadInto for AES-GCM and ChaCha20-Poly1305

Cel-Service:
**Relies on [this PR](https://github.com/RustCrypto/traits/pull/1663)!**

Resolves [this issue](https://github.com/RustCrypto/traits/issues/1311).

## Notes

I only implement `AeadInto` for AES-GCM and ChaChaPoly because these are the algorithms officially supported by the [Noise Protocol Framework](https://noiseprotocol.org/noise.html) (the driving factor behind my PR).

While I'm not personally invested in doing so, let me know if you need me to implement `AeadInto` for the other algorithms in order to accept this PR.

### DIFF
--- a/aes-gcm/src/lib.rs
+++ b/aes-gcm/src/lib.rs
@@ -108,7 +108,8 @@
 //! [`aead::Buffer`] for `arrayvec::ArrayVec` (re-exported from the [`aead`] crate as
 //! [`aead::arrayvec::ArrayVec`]).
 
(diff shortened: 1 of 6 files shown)


Cel-Service:
Addendum: It's perfectly acceptable to replace the pregenerated test data if so preferred. I only wanted non-zero, non-empty test data to confirm that encrypt_into and decrypt_into would work as expected with realistic parameters.

Cel-Service:
Retracted, as I am currently in progress writing the implementation for an alternative proposal.

PR #627, opened 2024-09-23
aead-stream: extract implementation from the aead crate

newpavlov:
This PR extracts implementation of the STREAM construction from the `aead` crate into a separate crate as was proposed in https://github.com/RustCrypto/traits/issues/1665.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -16,6 +16,13 @@ dependencies = [
 ]
 
 [[package]]
+name = "aead-stream"
+version = "0.1.0"
+dependencies = [
+ "aead",
+]
+
+[[package]]
 name = "aes"
 version = "0.9.0-pre.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -17,7 +17,7 @@ dependencies = [
 
 [[package]]
 name = "aead-stream"
-version = "0.1.0"
+version = "0.6.0-pre"
 dependencies = [
  "aead",
 ]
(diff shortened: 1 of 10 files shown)


aead-stream/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "aead-stream"
+version = "0.1.0"
```
tarcieri, line 3: Would suggest syncing the version with `aead` so it's easy to know which versions are compatible

```suggestion
version = "0.6.0-pre"
```

aead-stream/CHANGELOG.md
```diff
@@ -0,0 +1,8 @@
+# Changelog
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## 0.1.0 (2024-xx-xx)
```
tarcieri, line 7: Version number also here (maybe just leave it out for now?)

aead-stream/src/lib.rs
```diff
@@ -0,0 +1,229 @@
...
+    /// Compute the full AEAD nonce including the STREAM counter and last
+    /// block flag.
+    fn aead_nonce(&self, position: u32, last_block: bool) -> Result<aead::Nonce<A>> {
+        if position > Self::COUNTER_MAX {
+            return Err(Error);
+        }
+
+        let mut result = Array::default();
+
+        // TODO(tarcieri): use `generic_array::sequence::Concat` (or const generics)
```
tarcieri, line 220: Sidebar: should probably fix this, or at least update it to use `hybrid-array`

aead-stream/src/lib.rs
```diff
@@ -0,0 +1,229 @@
+#![no_std]
+#![doc = include_str!("../README.md")]
+#![allow(clippy::upper_case_acronyms)]
```
tarcieri, line 3: Might be good to fix this in a separate PR

Cargo.lock
```diff
@@ -308,9 +315,9 @@ checksum = "6fe2267d4ed49bc07b63801559be28c718ea06c4738b7a03c94df7386d2cde46"
 
 [[package]]
 name = "hybrid-array"
-version = "0.2.0-rc.9"
+version = "0.2.0-rc.10"
```
tarcieri, line 318: Might want to avoid this bump for now since it changes MSRV:

```suggestion
version = "0.2.0-rc.9"
```

newpavlov: Huh, I thought `hybrid-array` was pinned by upstream.

PR #641, opened 2024-10-23
Use full range of the stream counter for StreamLE31

Eosis:
I believe that we the `StreamPrimitive` for `StreamLE31` in the library is not using the correct value for `COUNTER_MAX`, choosing `0x0FFF_FFFF` rather than `0x7FFF_FFFF`, resulting in a reduced space for the message counter. This commit changes the value to use the full range.

I do not believe this is a security issue, but it does reduce the functionality of the library.

Let me know if my understanding of the problem is incorrect or if I'm attempting to commit this to the wrong library.

### DIFF
--- a/aead-stream/src/lib.rs
+++ b/aead-stream/src/lib.rs
@@ -177,7 +177,7 @@ where
     type NonceOverhead = U4;
     type Counter = u32;
     const COUNTER_INCR: u32 = 1;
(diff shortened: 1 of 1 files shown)


Eosis:
Sister PR in the crypto traits lib where I originally found this issue on the 0.5.2 release:
https://github.com/RustCrypto/traits/pull/1708

Issue #643, opened 2024-10-26
Chacha20poly1305 versions after and including 0.11.0-pre.1 require alloc on no_std.


ionspin:
`chacha20poly1305 = { version = "0.11.0-pre.1", default-features = false, features = [] }`

Build result:

```
error: no global memory allocator found but one is required; link to std or add `#[global_allocator]` to a static item that implements the GlobalAlloc trait
```

`chacha20poly1305 = { version = "0.10.1", default-features = false, features = [] }` builds successfuly.

I might be missing something obvious, it's late here :)

tarcieri:
My best guess is `chacha20poly1305` has a hard requirement for `chacha20/zeroize`, and it looks like `chacha20` is pulling `zeroize` in without `default-features = false`, which is linking liballoc

newpavlov:
>chacha20poly1305 has a hard requirement for chacha20/zeroize

But why is it a hard requirement? Most other crates have _optional_ support for zeroize.

>chacha20 is pulling zeroize in without default-features = false

Yeah, the problem is [here](https://github.com/RustCrypto/stream-ciphers/blob/master/chacha20/Cargo.toml#L26).

tarcieri:
It was likely introduced in recent changes. I’m on a phone so I would suggest using the git history to answer your question.

newpavlov:
Hm, it looks like `zeroize` was an unconditional dependency since v0.1 times. I guess it was an oversight, which did not get noticed later.

PR #644, opened 2024-10-27
Make zeroize optional for all crates

newpavlov:
See #643

### DIFF
--- a/aes-gcm-siv/src/lib.rs
+++ b/aes-gcm-siv/src/lib.rs
@@ -95,7 +95,6 @@ use cipher::{
     BlockCipherEncrypt, BlockSizeUser, InnerIvInit, StreamCipherCore,
 };
 use polyval::{universal_hash::UniversalHash, Polyval};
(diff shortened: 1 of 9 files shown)


aes-gcm-siv/src/lib.rs
```diff
@@ -261,9 +260,13 @@ where
...
         // TODO(tarcieri): use `Zeroizing` when const generics land
-        mac_key.as_mut_slice().zeroize();
-        enc_key.as_mut_slice().zeroize();
-        block.as_mut_slice().zeroize();
+        #[cfg(feature = "zeroize")]
+        {
+            use zeroize::Zeroize;
+            mac_key.as_mut_slice().zeroize();
+            enc_key.as_mut_slice().zeroize();
+            block.as_mut_slice().zeroize();
```
newpavlov, line 268: I am not sure we need these zeroizations of stack-based values since we generally do not deal with stack spilling.

tarcieri (approved):

PR #645, opened 2024-10-28
benches: fix compilation errors

### LINKED ISSUE #549: Consider using cycles for ascon-aead instead of milliseconds
All the other benchmarks in benches use cycles apart from ascon-aead. It will be great if it will also be able to measure it in cycles rather than in milliseconds.

newpavlov:
Also fix benches CI job and tweak ascon-aead benchmarks to make them more consistent with other benchmarks.

Closes #549

### DIFF
--- a/.github/workflows/benches.yml
+++ b/.github/workflows/benches.yml
@@ -23,11 +23,11 @@ jobs:
     strategy:
       matrix:
         rust:
-          - 1.65.0 # MSRV
+          - 1.81.0 # MSRV
           - stable
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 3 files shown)


tarcieri:
Hmm, `benches.yml` should probably run whenever any of the relevant crates are changed, or we should factor it into the individual crate-related `.yml` files and only run focused benchmarks for the one(s) that is/are changed.

It seems like because we don't have something like that, these breakages weren't spotted along with the breaking changes.

newpavlov:
I agree that we should run it on changes in the relevant crates, but it's not the reason why the breakage was not caught. It was because `cargo build --release` does not build benches.

PR #652, opened 2025-01-23
Add HS1-SIV

Demindiro:
HS1-SIV uses ChaCha and a new hash algorithm. This implementation is based on [the paper](https://krovetz.net/csus/papers/hs1-siv_v2.2.pdf) and the [reference implementation](https://bench.cr.yp.to/supercop.html).

I generated custom test vectors since none seem to be provided. I've included the reference implementation to show how they've been generated.

`Hs1Params` is quite ugly, but I'm unsure if I can make it any cleaner. It could be hidden by using newtypes for the 3 parameter sets instead.

`trait ChaChaImpl` is necessary because `chacha20::variants::Variant` and in particular `chacha20::variants::Ietf` isn't exposed, so `chacha20::ChaChaCore` is unusable.

I've spent some time optimizing it. It certainly can be optimized more, though so far further attempts have failed.

It should be free of any data-dependent branches, though I haven't looked at the generated assembly very closely.


Demindiro:
It's not clear to me why benches fails, since I can run it fine locally. Compiler version difference? (I use `rustc 1.86.0-nightly (8361aef0d 2025-01-14)`)

newpavlov:
I don't think we need the `ref` folder for anything. It does not look like it's used for tests or something like this. Or am I missing something?

>Compiler version difference?

You could try bumping the compiler version used for benchmarks.

Demindiro:
I believe the failing mgm test isn't related to this PR?

tarcieri:
@Demindiro it does look unrelated, yes

newpavlov:
The `mgm` CI failure is fixed in #654, so you need to rebase.

PR #664, opened 2025-03-14
finish inout migration

baloo:


tarcieri:
Nice, thanks

deoxys/Cargo.toml
```diff
@@ -35,6 +35,7 @@ bytes = ["aead/bytes"]
 os_rng  = ["aead/os_rng", "rand_core"]
 heapless = ["aead/heapless"]
 rand_core = ["aead/rand_core"]
+inout = ["aead/inout"]
```
tarcieri, line 38: Why are you making this a feature? It's already a hard dependency of `aead`

baloo: I started this way before I realized it didn't make sense, that shouldn't have gone through, sorry.

aes-gcm/src/lib.rs
```diff
@@ -301,11 +301,11 @@ where
 
         // TODO(tarcieri): interleave encryption with GHASH
         // See: <https://github.com/RustCrypto/AEADs/issues/74>
-        let expected_tag = self.compute_tag(mask, associated_data, buffer);
+        let expected_tag = self.compute_tag(mask, associated_data, buffer.get_out());
```
tarcieri, line 304: Shouldn't this be `get_in`?

aes-gcm-siv/src/lib.rs
```diff
@@ -301,8 +301,8 @@ where
         self.polyval.update_padded(associated_data);
 
         // TODO(tarcieri): interleave decryption and authentication
-        init_ctr(&self.enc_cipher, tag).apply_keystream_partial(buffer.into());
-        self.polyval.update_padded(buffer);
+        init_ctr(&self.enc_cipher, tag).apply_keystream_partial(buffer.reborrow());
+        self.polyval.update_padded(buffer.get_in());
```
tarcieri, line 305: I think this one needs to be `get_out`, since it's authenticating the decrypted plaintext (since the "SIV" tag is calculated from the plaintext)

aes-gcm/src/lib.rs
```diff
@@ -280,9 +280,9 @@ where
 
         // TODO(tarcieri): interleave encryption with GHASH
         // See: <https://github.com/RustCrypto/AEADs/issues/74>
-        ctr.apply_keystream_partial(buffer.into());
+        ctr.apply_keystream_partial(buffer.reborrow());
 
-        let full_tag = self.compute_tag(mask, associated_data, buffer);
+        let full_tag = self.compute_tag(mask, associated_data, buffer.get_in());
```
tarcieri, line 285: This one probably needs to be `get_out`, to compute a MAC over the ciphertext

baloo: Yeah there is probably a whole slew of those, I assumed the backing buffer was unique.

aes-siv/src/siv.rs
```diff
@@ -227,15 +227,15 @@ where
...
     ) -> Result<Tag, Error>
     where
         I: IntoIterator<Item = T>,
         T: AsRef<[u8]>,
     {
         // Compute the synthetic IV for this plaintext
-        let siv_tag = s2v(&mut self.mac, headers, plaintext)?;
-        self.xor_with_keystream(siv_tag, plaintext);
+        let siv_tag = s2v(&mut self.mac, headers, plaintext.get_in())?;
+        self.xor_with_keystream(siv_tag, plaintext.get_out());
```
tarcieri, line 238: This probably needs to pass `plaintext` and the receiver needs to load plaintext from in and write to out

aes-gcm-siv/src/lib.rs
tarcieri: Yes

baloo:
Need to remake this PR from scratch.

tarcieri:
@baloo you can just open a new complete PR to `master` if you want and I can close #663, that would make for a simpler history

baloo:
I think I need a util for inout to duplicate the backing buffer from the original slice to flush out the misuse of `get_in()`

tarcieri:
@baloo that's what I was suggesting re: smoke tests, which can use separate buffers for input and output. Once you have a basic framework for them you can put them in `aead::dev`

PR #665, opened 2025-03-14
Migrate to `AeadInOut`

baloo:
Replaces the previous `AeadInPlaceDetached` impls with `AeadInOut`, which was introduced in RustCrypto/traits#1793

Depends:
 - #666 
 - #667
 - #673 
 - https://github.com/RustCrypto/traits/pull/1797


aes-gcm/tests/common/mod.rs
```diff
@@ -35,6 +35,26 @@ macro_rules! tests {
             }
         }
 
+        #[test]
+        fn encrypt_inout() {
+            for vector in $vectors {
+                let key = Array(*vector.key);
+                let nonce = Array(*vector.nonce);
+                let mut payload = BicephalBuffer::from(vector.plaintext);
```
baloo, line 43: This makes use of https://github.com/RustCrypto/traits/pull/1797

newpavlov:
I would wait with this migration a bit. I plan to open several PRs to change `aead` APIs further and we probably will need  to discuss them a fair amount.

baloo:
I don't mind waiting to merge but I'll get this branch to a point where it can be used in a `[patch.crates-io]` by a consumer which is not currently the case with `master`.

tarcieri:
It would definitely be nice to get things in sync with `master` if we can. There are people who are trying to track these repos as git dependencies and that becomes difficult when they go out-of-sync. Ideally we'd have a PR to update the repo before we merge changes in `traits`, which is my bad

baloo:
I'll need to sort out the lack of `MockBuffer` for ascon and deoxys: https://github.com/RustCrypto/traits/pull/1802#discussion_r2008056556

tarcieri:
@baloo can you update/recreate #673 so we can get this merged?

baloo:
> @baloo can you update/recreate #673 so we can get this merged?

https://github.com/RustCrypto/AEADs/pull/687

This gets included here.

baloo:
I think this is the last PR needed to reunite the workspace in formats.

tarcieri (approved):

PR #666, opened 2025-03-16
deoxys: rewrite the internals to use inout

baloo:
This is to prepare the migration to `AeadInOut`, following https://github.com/RustCrypto/traits/pull/1793

Note: there is a strong chance this could actually use the StreamCipherCore api, but I couldn't not make it fit.

### DIFF
--- a/deoxys/Cargo.toml
+++ b/deoxys/Cargo.toml
@@ -20,6 +20,7 @@ rust-version = "1.85"
 [dependencies]
 aead = { version = "0.6.0-rc.0", default-features = false }
 aes = { version = "=0.9.0-pre.3", features = ["hazmat"], default-features = false }
+inout = { version = "0.2.0-rc.4", default-features = false }
 subtle = { version = "2", default-features = false }
 zeroize = { version = "1", optional = true, default-features = false }
 
--- a/deoxys/src/lib.rs
+++ b/deoxys/src/lib.rs
@@ -324,3 +324,114 @@ where
     B: DeoxysBcType,
 {
 }
+
+#[cfg(test)]
+mod tests {
+    //! this module is here to test the inout behavior which is not currently exposed.
+    //! it will be once we port over to the API made in RustCrypto/traits#1793.
+    //!
+    //! This is to drop once https://github.com/RustCrypto/traits/pull/1797 is made available.
+    //!
+    //! It duplicates test vectors from `tests/deoxys_i_128.rs` and provides a mock buffer backing
+    //! for InOut.
+
+    use hex_literal::hex;
+
+    use super::*;
+
+    struct MockBuffer {
+        in_buf: [u8; 33],
+        out_buf: [u8; 33],
+    }
+
+    impl From<&[u8]> for MockBuffer {
+        fn from(buf: &[u8]) -> Self {
+            let mut in_buf = [0u8; 33];
+            in_buf.copy_from_slice(buf);
+            Self {
+                in_buf,
+                out_buf: [0u8; 33],
+            }
+        }
+    }
+
+    impl MockBuffer {
+        /// Get an [`InOutBuf`] from a [`MockBuffer`]
+        pub fn to_in_out_buf(&mut self) -> InOutBuf<'_, '_, u8> {
+            InOutBuf::new(self.in_buf.as_slice(), self.out_buf.as_mut_slice())
+                .expect("Invariant violation")
+        }
+    }
+
+    impl AsRef<[u8]> for MockBuffer {
+        fn as_ref(&self) -> &[u8] {
+            &self.out_buf
+        }
+    }
+
+    #[test]
+    fn test_deoxys_i_128_5() {
+        let plaintext = hex!("5a4c652cb880808707230679224b11799b5883431292973215e9bd03cf3bc32fe4");
+        let mut buffer = MockBuffer::from(&plaintext[..]);
+
+        let aad = [];
+
+        let key = hex!("101112131415161718191a1b1c1d1e1f");
+        let key = Array(key);
+
+        let nonce = hex!("202122232425262728292a2b2c2d2e2f");
+        let nonce = Array::try_from(&nonce[..8]).unwrap();
+
+        let ciphertext_expected =
+            hex!("cded5a43d3c76e942277c2a1517530ad66037897c985305ede345903ed7585a626");
+
+        let tag_expected: [u8; 16] = hex!("cbf5faa6b8398c47f4278d2019161776");
+
+        type M = modes::DeoxysI<deoxys_bc::DeoxysBc256>;
+        let cipher = DeoxysI128::new(&key);
+        let tag: Tag = M::encrypt_inout(&nonce, &aad, buffer.to_in_out_buf(), &cipher.subkeys);
+
+        let ciphertext = buffer.as_ref();
+        assert_eq!(ciphertext, ciphertext_expected);
+        assert_eq!(tag, tag_expected);
+
+        let mut buffer = MockBuffer::from(buffer.as_ref());
+        M::decrypt_inout(&nonce, &aad, buffer.to_in_out_buf(), &tag, &cipher.subkeys)
+            .expect("decryption failed");
+
+        assert_eq!(&plaintext[..], buffer.as_ref());
+    }
+
+    #[test]
+    fn test_deoxys_ii_128_5() {
+        let plaintext = hex!("06ac1756eccece62bd743fa80c299f7baa3872b556130f52265919494bdc136db3");
+        let mut buffer = MockBuffer::from(&plaintext[..]);
+
+        let aad = [];
+
+        let key = hex!("101112131415161718191a1b1c1d1e1f");
+        let key = Array(key);
+
+        let nonce = hex!("202122232425262728292a2b2c2d2e2f");
+        let nonce = Array::try_from(&nonce[..15]).unwrap();
+
+        let ciphertext_expected =
+            hex!("82bf241958b324ed053555d23315d3cc20935527fc970ff34a9f521a95e302136d");
+
+        let tag_expected: [u8; 16] = hex!("0eadc8612d5208c491e93005195e9769");
+
+        type M = modes::DeoxysII<deoxys_bc::DeoxysBc256>;
+        let cipher = DeoxysII128::new(&key);
+        let tag: Tag = M::encrypt_inout(&nonce, &aad, buffer.to_in_out_buf(), &cipher.subkeys);
+
+        let ciphertext = buffer.as_ref();
+        assert_eq!(ciphertext, ciphertext_expected);
+        assert_eq!(tag, tag_expected);
+
+        let mut buffer = MockBuffer::from(buffer.as_ref());
+        M::decrypt_inout(&nonce, &aad, buffer.to_in_out_buf(), &tag, &cipher.subkeys)
+            .expect("decryption failed");
+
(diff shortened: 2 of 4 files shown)


deoxys/Cargo.toml
```diff
@@ -20,6 +20,7 @@ rust-version = "1.85"
 [dependencies]
 aead = { version = "0.6.0-rc.0", default-features = false }
 aes = { version = "=0.9.0-pre.3", features = ["hazmat"], default-features = false }
+inout = { version = "0.2.0-rc.4", default-features = false }
```
baloo, line 23: Eventually this should drop and use `aead::inout`

tarcieri:
@baloo can you add one or more tests that actually exercise `InOutBuf` with separate backing buffers?

baloo:
Yeah, found a misuse of `get_in` haha ><

deoxys/src/lib.rs
```diff
@@ -321,3 +324,115 @@ where
     B: DeoxysBcType,
 {
 }
+
+#[cfg(all(feature = "alloc", test))]
```
tarcieri, line 328: If you put these tests in the toplevel `tests/` directory they automatically get std/alloc

baloo: yeah but I can't put them in tests I don't have access to the subkeys there.

tarcieri (approved):

deoxys/src/lib.rs
baloo: Update this is now the last commit in https://github.com/RustCrypto/AEADs/pull/665

Cargo.toml
```diff
@@ -11,6 +11,8 @@ members = [
     "eax",
     "ocb3",
     "xaes-256-gcm",
+
+    "benches",
```
newpavlov, line 15: `benches` are intentionally excluded from the workspace.

deoxys/src/modes.rs
```diff
@@ -327,38 +349,49 @@ where
...
+                $xor(&block);
 
                 // XOR out block numbers
                 for (t, i) in tweak[8..].iter_mut().zip(&index_array) {
                     *t ^= i
                 }
+            };
+        }
+
+        if !buffer.is_empty() {
```
newpavlov, line 381: To prevent drift to the right, it's worth to write this instead:
```rust
if buffer.is_empty() {
    return;
}
```

Same in the `authenticate_message` method.

deoxys/src/modes.rs
```diff
@@ -327,38 +349,49 @@ where
...
         nonce: &Array<u8, U15>,
     ) {
-        if !buffer.is_empty() {
-            tweak.copy_from_slice(tag);
-            tweak[0] |= 0x80;
-
-            for (index, data) in buffer.chunks_mut(16).enumerate() {
-                let index_array = (index as u64).to_be_bytes();
+        macro_rules! encrypt_decrypt_block {
+            ($index: ident, $xor: expr) => {
```
newpavlov, line 359: I think we can use a function (defined inside of the method) here instead:
```rust
fn encrypt_decrypt_block(index: usize, xor: impl FnOnce(block: &Block)) { ... }
```

deoxys/src/modes.rs
```diff
@@ -327,38 +349,49 @@ where
...
+        }
+
+        if !buffer.is_empty() {
+            tweak.copy_from_slice(tag);
+            tweak[0] |= 0x80;
+
+            let (blocks, tail) = buffer.into_chunks::<U16>();
+            let blocks_len = blocks.len();
+            for (index, mut data) in blocks.into_iter().enumerate() {
+                encrypt_decrypt_block!(index, |block: &Array<u8, U16>| data.xor_in2out(block));
```
newpavlov, line 388: Why not use the `Block` alias?

deoxys/src/modes.rs
```diff
@@ -299,10 +321,10 @@ where
                 tweak[8..].copy_from_slice(&(index as u64).to_be_bytes());
 
                 if data.len() == 16 {
-                    let mut block = [0u8; 16];
+                    let mut block = Block::default();
```
newpavlov, line 324: For consistency I think it's worth to change this method to use `chunks_exact` and split whole blocks and tail processing similarly to `encrypt_decrypt_message`.

newpavlov (approved):
This PR includes a fair amount of changes which are not directly relevant to the `inout` support. Ideally, I would prefer to have them in a separate PR, but it's fine to merge it as-is, if you don't wan to bother with the split.

deoxys/src/lib.rs
```diff
@@ -139,6 +140,12 @@ pub type Nonce<NonceSize> = Array<u8, NonceSize>;
 /// Deoxys tags
 pub type Tag = Array<u8, U16>;
 
+type Block = Array<u8, U16>;
+
+type Tweak = Array<u8, U16>;
+
+type DeoxysKey = Array<u8, U16>;
```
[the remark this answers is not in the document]
baloo: It conflicted with `cipher::Key`, got lazy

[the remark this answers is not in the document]
baloo: `Key` is re-exported from `cipher` from the top level. I can't redefine the `Key` while also exporting it.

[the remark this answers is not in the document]
newpavlov: You can remove this re-export and use this alias instead of `Key<Self>`.

deoxys/src/lib.rs
```diff
@@ -321,3 +324,115 @@ where
...
+    impl AsRef<[u8]> for MockBuffer {
+        fn as_ref(&self) -> &[u8] {
+            &self.out_buf
+        }
+    }
+
+    #[test]
+    fn test_deoxys_i_128_5() {
+        let plaintext = hex!("5a4c652cb880808707230679224b11799b5883431292973215e9bd03cf3bc32fe4");
+        let mut buffer = MockBuffer::from(&plaintext[..]);
```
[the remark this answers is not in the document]
baloo: I still need the alloc for the inner Vec of MockBuffer

Cargo.toml
newpavlov: You could either use a concrete git commit of the traits repo (before the inout changes) in the benches patch section, or you could temporarily disable the benches CI job.

deoxys/src/lib.rs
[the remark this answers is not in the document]
newpavlov: My point is that you do not need heap-allocated `MockBuffer` for these tests and can use stack-allocated buffers, same for #665.

deoxys/src/lib.rs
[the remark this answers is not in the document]
newpavlov: Ah, I missed that `DeoxysKey` is a private alias. I thought it was public.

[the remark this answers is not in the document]
newpavlov: Since Dyoxis uses only one key size, I think we can export only `pub type Key = Array<u8; u16>;` since I am not sure that the generic alias is that useful. In other words, the question whether users will prefer to use `deoxys::Key` or `Key<DeoxysI128>`.

@tarcieri WDYT?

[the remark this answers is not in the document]
tarcieri: It seems fine to me to export the alias to `Array<u8; 16>`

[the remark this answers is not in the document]
baloo: https://github.com/RustCrypto/AEADs/pull/672

baloo:
I think this one is ready to merge, unless you have anything other to say.

deoxys/src/modes.rs
```diff
@@ -112,66 +114,75 @@ where
...
-                    // Last block checksum
-                    tweak[0] = (tweak[0] & 0xf) | TWEAK_M_LAST;
+                B::encrypt_inout(data, &tweak, subkeys);
+            }
 
-                    let mut block = Block::default();
-                    block[0..data.len()].copy_from_slice(data);
+            let mut data = tail;
+            let index = data_blocks_len;
+            if !data.is_empty() {
```
newpavlov, line 135: `buffer_len % 16 == 0` is always true when `!data.is_empty()` is false. I think it would be better to write this part like this:
```rust
if tail.is_empty() {
    // Without incomplete last block
    ...
} else {
    // With incomplete last block
    ...
}
```

Same for `decrypt_inout`.

baloo: the best I can do is will look this:
```
diff --git a/deoxys/src/modes.rs b/deoxys/src/modes.rs
index 0f91ec36e0..e6ed2cd000 100644
--- a/deoxys/src/modes.rs
+++ b/deoxys/src/modes.rs
@@ -111,12 +111,11 @@
         tweak[8] = nonce[7] << 4;

         // Message authentication and encryption
-        if !buffer.is_empty() {
+        let (data_blocks, mut tail) = buffer.into_chunks();
+        let data_blocks_len = data_blocks.len();
+        if !data_blocks.is_empty() || !tail.is_empty() {
             tweak[0] = (tweak[0] & 0xf) | TWEAK_M;

-            let (data_blocks, tail) = buffer.into_chunks();
-            let data_blocks_len = data_blocks.len();
-
             for (index, data) in data_blocks.into_iter().enumerate() {
                 // Copy block number
                 let tmp = tweak[8] & 0xf0;
@@ -130,9 +129,8 @@
                 B::encrypt_inout(data, &tweak, subkeys);
             }

-            let mut data = tail;
             let index = data_blocks_len;
-            if !data.is_empty() {
+            if !tail.is_empty() {
                 // Last block, incomplete

                 // Copy block number
@@ -144,9 +142,9 @@
                 tweak[0] = (tweak[0] & 0xf) | TWEAK_M_LAST;

                 let mut block = Block::default();
-                block[0..data.len()].copy_from_slice(data.get_in());
+                block[0..tail.len()].copy_from_slice(tail.get_in());

-                block[data.len()] = 0x80;
+                block[tail.len()] = 0x80;

                 for (c, d) in checksum.iter_mut().zip(block.iter()) {
                     *c ^= d;
@@ -157,24 +155,23 @@
                 // Last block encryption
                 B::encrypt_inout((&mut block).into(), &tweak, subkeys);

-                data.xor_in2out((block[..data.len()]).into());
-
-                // Tag computing.
-                tweak[0] = (tweak[0] & 0xf) | TWEAK_CHKSUM;
-
-                let tmp = tweak[8] & 0xf0;
-                tweak[8..].copy_from_slice(&((index + 1) as u64).to_be_bytes());
-                tweak[8] = (tweak[8] & 0xf) | tmp;
-
-                B::encrypt_inout((&mut checksum).into(), tweak.as_ref(), subkeys);
-
-                for (t, c) in tag.iter_mut().zip(checksum.iter()) {
-                    *t ^= c;
-                }
+                tail.xor_in2out((block[..tail.len()]).into());
             }
         }

-        if buffer_len % 16 == 0 {
+        if !tail.is_empty() {
+            tweak[0] = (tweak[0] & 0xf) | TWEAK_CHKSUM;
+
+            let tmp = tweak[8] & 0xf0;
+            tweak[8..].copy_from_slice(&((data_blocks_len + 1) as u64).to_be_bytes());
+            tweak[8] = (tweak[8] & 0xf) | tmp;
+
+            B::encrypt_inout((&mut checksum).into(), tweak.as_ref(), subkeys);
+
+            for (t, c) in tag.iter_mut().zip(checksum.iter()) {
+                *t ^= c;
+            }
+        } else {
             // Tag computing without last block
             tweak[0] = (tweak[0] & 0xf) | TWEAK_TAG;
 ```
 
 That makes the whole patch series a lot less readable in my opinion (even when ignoring whitespace).

PR #667, opened 2025-03-17
ocb3: migrate internals to use inout

baloo:
This is to prepare the migration to `AeadInOut`, following https://github.com/RustCrypto/traits/pull/1793

Depends:
 - https://github.com/RustCrypto/utils/pull/1170

### DIFF
--- a/ocb3/src/lib.rs
+++ b/ocb3/src/lib.rs
@@ -21,11 +21,12 @@ pub use aead::{
 use aead::{PostfixTagged, array::ArraySize};
 use cipher::{
     BlockCipherDecrypt, BlockCipherEncrypt, BlockSizeUser,
-    consts::{U12, U16},
-    typenum::Unsigned,
+    consts::{U2, U12, U16},
+    typenum::Prod,
 };
 use core::marker::PhantomData;
 use dbl::Dbl;
+use inout::{InOut, InOutBuf};
 use subtle::ConstantTimeEq;
 
 /// Number of L values to be precomputed. Precomputing m values, allows
@@ -55,7 +56,9 @@ pub type Nonce<NonceSize> = Array<u8, NonceSize>;
 /// OCB3 tag
 pub type Tag<TagSize> = Array<u8, TagSize>;
 
-pub(crate) type Block = Array<u8, U16>;
+type BlockSize = U16;
+pub(crate) type Block = Array<u8, BlockSize>;
+type DoubleBlock = Array<u8, Prod<BlockSize, U2>>;
 
 mod sealed {
     use aead::array::{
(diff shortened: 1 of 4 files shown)


ocb3/src/lib.rs
```diff
@@ -580,11 +590,20 @@ pub(crate) fn ntz(n: usize) -> usize {
...
+) -> [InOut<'i, 'o, Block>; 2] {
+    let (input, output) = two_blocks.into_raw();
+
+    let (bi0, bi1) = unsafe { input.as_ref() }
+        .unwrap()
+        .split_at(BlockSize::USIZE);
+    let (bi0, bi1): (&Block, &Block) = (bi0.try_into().unwrap(), bi1.try_into().unwrap());
+    let (bo0, bo1) = unsafe { output.as_mut() }
+        .unwrap()
+        .split_at_mut(BlockSize::USIZE);
```
baloo, line 604: Those two unsafe blocks are less than ideal obviously, but I couldn't figure out how to make InOut split the array otherwise.

baloo:
I couldn't figure out a way to test inout here, but this gets tested in https://github.com/RustCrypto/AEADs/pull/665 when it's plugged with `AeadInOut` (tested by `aead::new_test`)

ocb3/src/lib.rs
baloo: something ala https://docs.rs/inout/0.2.0-rc.4/inout/struct.InOutBuf.html#method.split_at

baloo: @newpavlov any opinion on this one?

ocb3/src/lib.rs
```diff
@@ -55,7 +56,9 @@ pub type Nonce<NonceSize> = Array<u8, NonceSize>;
 /// OCB3 tag
 pub type Tag<TagSize> = Array<u8, TagSize>;
 
-pub(crate) type Block = Array<u8, U16>;
+type BlockSize = U16;
+pub(crate) type Block = Array<u8, BlockSize>;
+type DoubleBlock = Array<u8, <BlockSize as Mul<U2>>::Output>;
```
tarcieri, line 61: Alternatively you can use `Add` instead of `Mul<U2>` (where `Add` implicitly takes `Self` as a default) e.g. https://github.com/RustCrypto/AEADs/blob/a09f3d2/aes-siv/src/siv.rs#L101

baloo: It's an internal type, I think it's better to be explicit about the intent?
Relying on the implicit value of Add makes it harder to read imho.

baloo: went with `typenum::Prod`

tarcieri (approved):

PR #671, opened 2025-03-17
deoxys: reorganize `authenticate_message`

baloo:
This is split off from https://github.com/RustCrypto/AEADs/pull/666

### DIFF
--- a/deoxys/src/modes.rs
+++ b/deoxys/src/modes.rs
@@ -294,38 +294,46 @@ where
         subkeys: &Array<DeoxysKey, B::SubkeysSize>,
         tag: &mut Tag,
     ) {
-        if !buffer.is_empty() {
-            tweak[0] = TWEAK_M;
+        if buffer.is_empty() {
+            return;
+        }
+        tweak[0] = TWEAK_M;
 
-            for (index, data) in buffer.chunks(16).enumerate() {
-                // Copy block number
-                tweak[8..].copy_from_slice(&(index as u64).to_be_bytes());
+        let (chunks, tail) = Block::slice_as_chunks(buffer);
 
-                if data.len() == 16 {
-                    let mut block = Block::default();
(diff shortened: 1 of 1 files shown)


deoxys/src/modes.rs
```diff
@@ -294,38 +294,49 @@ where
...
-            tweak[0] = TWEAK_M;
+        if buffer.is_empty() {
+            return;
+        }
+        tweak[0] = TWEAK_M;
 
-            for (index, data) in buffer.chunks(16).enumerate() {
-                // Copy block number
-                tweak[8..].copy_from_slice(&(index as u64).to_be_bytes());
+        let mut chunks = buffer.chunks_exact(16);
```
tarcieri, line 302: You could use [`Array::slice_as_chunks`](https://docs.rs/hybrid-array/latest/hybrid_array/struct.Array.html#method.slice_as_chunks) here

tarcieri (approved):

deoxys/src/modes.rs
```diff
@@ -294,38 +294,47 @@ where
...
-            for (index, data) in buffer.chunks(16).enumerate() {
-                // Copy block number
-                tweak[8..].copy_from_slice(&(index as u64).to_be_bytes());
+        let (chunks, tail) = Block::slice_as_chunks(buffer);
+        let chunks_len = chunks.len();
 
-                if data.len() == 16 {
-                    let mut block = Block::default();
-                    block.copy_from_slice(data);
+        for (index, data) in chunks.into_iter().enumerate() {
```
tarcieri, line 305: Hit approve then saw the clippy warning. It seems like if you use `chunks.iter()` rather than `chunks.into_iter()`, it won't consume `chunks`, and then you can just use `chunks.len()`

tarcieri (approved):

PR #672, opened 2025-03-17
deoxys: export `Key` type alias instead of `aead::Key`

baloo:


tarcieri:
Hmm, now I wonder if this will be confusing with `Deoxys*256` which uses 32-byte keys

baloo:
uh, right, yeah I'll just keep the `DeoxysKey`

PR #673, opened 2025-03-17
ascon-aead: rewrite internals to use inout

baloo:
part of #665

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -99,6 +99,7 @@ dependencies = [
  "aead",
  "ascon",
  "hex-literal",
+ "inout",
  "subtle",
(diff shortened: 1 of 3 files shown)


sebastinas:
What is the benefit of this? How does it compare in overhead to u8 slices?

newpavlov:
I allows us to use disjoint buffers for plaintext and ciphertext, which allows us to remove unnecessary copies in some cases. As for overhead, in the worst case compiler would have to keep one additional destination pointer, but in most cases compiler should see that source and destination pointers are the same.

sebastinas:
ascon-aead will be maintained elsewhere going forward.

PR #675, opened 2025-03-22
ascon-aead: migrate tests to blobby

newpavlov:
The new test uses blobby format which will be used by the new `new_pass_test!` macro (see https://github.com/RustCrypto/traits/pull/1803).


newpavlov:
@sebastinas 
Do you remember provenance of the KATs?

It also looks like these [Wycheproof vectors](https://github.com/C2SP/wycheproof/blob/60744a36/testvectors_v1/ascon128_test.json) are wrong? For `key: "b67b1a6efdd40d37080fbe8f8047aeb9"`, `"iv": "fa294b129972f7fc5bbd5b96bba837c9"`, `"aad" : ""`, and `"msg" : ""` from our implementation I get tag equal to `dc1acf9d5a18658c39b1337c0e02c07d`, while the first Wycheproof vector expects `47648fcad24982437276b8d5901f812b`.

newpavlov:
Regenerated tests based using [vectors](https://github.com/ascon/ascon-c/blob/fdfca408/crypto_aead/asconaead128/LWC_AEAD_KAT_128_128.txt) from the reference implementation.

sebastinas:
If you have questions, please at least give me time to answer them before you merge.

PR #676, opened 2025-03-23
ascon-aead: remove license headers from source files

newpavlov:
We generally do not include license headers in source files and rely solely on the LICENSE files instead.

### DIFF
--- a/ascon-aead/src/asconcore.rs
+++ b/ascon-aead/src/asconcore.rs
@@ -1,6 +1,3 @@
-// Copyright 2021-2023 Sebastian Ramacher
-// SPDX-License-Identifier: Apache-2.0 OR MIT
-
 use aead::{
     Error,
     array::{Array, ArraySize},
--- a/ascon-aead/src/lib.rs
+++ b/ascon-aead/src/lib.rs
@@ -1,6 +1,3 @@
-// Copyright 2021-2023 Sebastian Ramacher
-// SPDX-License-Identifier: Apache-2.0 OR MIT
-
 #![no_std]
 #![cfg_attr(docsrs, feature(doc_auto_cfg))]
 #![doc = include_str!("../README.md")]


sebastinas:
I have removed upload rights for ascon-aead and ascon-hashes until trust is restored.

newpavlov:
It's a matter of consistency across our crates. Your original authorship is kept in [LICENSE-MIT](https://github.com/RustCrypto/AEADs/blob/master/ascon-aead/LICENSE-MIT#L1) and even in the [`authors`](https://github.com/RustCrypto/AEADs/blob/master/ascon-aead/Cargo.toml#L6) (though we usually just use "RustCrypto Developers" in almost all our other crates).

Why duplicating the licensing information in each source file is important for you? We do not use different licensing in different files like in some messy old projects of yore and tracking each individual contribution for each file would become incredibly messy quite quickly. I can revert this change if it's _that_ important for you for some reason, but I believe it would stuck out as a sore thumb.

sebastinas:
I will maintain it elsewhere then. thx and bye

newpavlov:
I do not believe I did anything wrong other than failing to mention it in the original PR.

We can continue development under the `ascon-aead128` name without any strings attached.

tarcieri:
I'd agree with @baloo here: this isn't worth fracturing the crate ecosystem over. Let's revert and open an issue to discuss.

newpavlov:
As I wrote above, I am open to reverting this change despite my dislike license headers in each license file. But it seems @sebastinas has made the decision without bothering to explain why they are so important for him.

newpavlov:
Tangential note: in future we should demand a full transfer of crate ownership. Otherwise, we have the risk of original author releasing new crate versions independently from source code state in our repos (either intentionally, or because of compromised account) or making some demands which go against our established development practices.

sebastinas:
> As I wrote above, I am open to reverting this change despite my dislike for license headers in each source file. But it seems @sebastinas has made the decision without bothering to explain why they are so important for him.

You never ever remove license information by other authors without prior consent. That's just basic respect between open source maintainers. If you like them or not doesn't make a difference. What's next? That you remove/change Copyright claims because they are not consistent with other crates?

newpavlov:
>You never ever remove license information by other authors without prior consent.

I did not remove the licensing information from the LICENSE-MIT file and do not intend to. It's the place which we and many other Rust projects use to keep licensing information. For example, look at the `std` code, **none** of its source files contains license headers. Each project has its own development rules and consistency of our crates is one of their strengths and we are fully in the legal field here since we include the licensing information in a "substantial portion" of the software in the form of the crate.

>What's next? That you remove/change Copyright claims because they are not consistent with other crates?

No. As you can see in other our crates we keep original contributors in LICENSE files. In the oldest crates we even mention Graydon Hoare, despite the fact that probably none of his original code is left in our crates.

tarcieri:
@newpavlov I think the issue here is one of consent, that you removed the comments before getting @sebastinas's sign-off, regardless of what the motivation is

newpavlov:
And what if there is no consent? What if Sebastian is against migration to `inout` as he alluded in the other comment? What about potential future refactorings/optimizations? Should we acquire his consent for any future change? This is what it means to have "strings attached" and I would prefer to develop crate under a different name than to deal with this.

newpavlov:
>I don't think it's too much to ask to get @sebastinas's input on changes.

I am open to his input, but I am not open to his _dictate_ of the crate development.

>this is specifically an issue with removing authorship information without consent

Again, I do not believe I "removed authorship". We have a dedicated place for original/substantial contributors in the license files (and in some extremely rare case the `authors` field), everything else is inconsistency.

sebastinas:
> And what if there is no consent? What if Sebastian is against migration to `inout` as he alluded in the other comment? What about potential future refactorings/optimizations? Should we acquire his consent for any future change? This is what it means to have "strings attached" and I would prefer to develop crate under a different name than to deal with this.

Then you misrepresent my statement here. I have asked what the benefit is. I only closed the MR after announcing that I will remove ascon-aead from RustCrypto.

But if you think that by me asking for obtaining for consent before removing license comments from files I am dictating development practices without ever blocking technical changes, then it's best for everyone if we part ways.

tarcieri:
@sebastinas is there anything we can do to fix the situation here?

PR #677, opened 2025-03-23
Rename `ascon-aead` to `ascon-aead128`

newpavlov:


tarcieri:
This seems way too hasty, how about reverting the change that instigated this incident as was [suggested here](https://github.com/RustCrypto/AEADs/pull/676#issuecomment-2746480695)?

newpavlov:
I do not plan to merge it right away and plan to wait wait at least several days.

baloo:
Could we close this PR and shelf it for now?

baloo:
You don't go try to make peace with a loaded gun. Closing the PR doesn't stop you from re-opening it later.

newpavlov:
I place the "loaded gun" on the table after the shots were [fired](https://github.com/RustCrypto/AEADs/pull/676#issuecomment-2746453489). And if by "making peace" you mean "giving in to the blackmail threats", then I would prefer to just use a different name.

I still did not get a proper explanation why the license headers are _that_ important and why Sebastian should get a special treatment, while all other contributors (me and you included) continue to work on the project just fine without license headers with their name in each source file touched or created by them.

PR #678, opened 2025-03-23
Revert "ascon-aead: remove license headers from source files (#676)"

tarcieri:
This reverts commit aafc20ed4af21c7c2246e10ce80147a5ad1196d5.

See discussion on #676


tarcieri:
@sebastinas if this looks good to you, it would be great to figure out a solution that doesn't involve forking/fragmenting the crate ecosystem

newpavlov:
I still would like to hear the reason for _why_ it's important to have license headers in each file.

PR #680, opened 2025-03-24
deoxys: tweak `DeoxysMode::encrypt/decrypt_inout` methods

newpavlov:
The change reduces code drift to the right and simplifies the branching structure.

It was previously suggested [here](https://github.com/RustCrypto/AEADs/pull/666#discussion_r2009532388).

### DIFF
--- a/deoxys/src/modes.rs
+++ b/deoxys/src/modes.rs
@@ -92,7 +92,6 @@ where
         let mut tag = Tag::default();
         let mut checksum = Checksum::default();
         let mut tweak = Tweak::default();
-        let buffer_len = buffer.len();
 
         // Associated Data
         <Self as DeoxysModeInternal<B>>::compute_ad_tag(
@@ -115,7 +114,7 @@ where
         tweak[0] = (tweak[0] & 0xf) | TWEAK_M;
 
         let (data_blocks, mut tail) = buffer.into_chunks();
-        let data_blocks_len = data_blocks.len();
+        let mut data_blocks_len = data_blocks.len();
 
         for (index, data) in data_blocks.into_iter().enumerate() {
             // Copy block number
@@ -130,26 +129,11 @@ where
             B::encrypt_inout(data, &tweak, subkeys);
         }
 
-        if tail.is_empty() {
-            // Tag computing without last block
-            tweak[0] = (tweak[0] & 0xf) | TWEAK_TAG;
-
-            let tmp = tweak[8] & 0xf0;
-            tweak[8..].copy_from_slice(&((buffer_len / 16) as u64).to_be_bytes());
-            tweak[8] = (tweak[8] & 0xf) | tmp;
-
-            B::encrypt_inout((&mut checksum).into(), tweak.as_ref(), subkeys);
-
-            for (t, c) in tag.iter_mut().zip(checksum.iter()) {
-                *t ^= c;
-            }
-        } else {
-            // Tag computing with incomplete last block
-            let index = data_blocks_len;
-
+        // Process incomplete last block
+        if !tail.is_empty() {
             // Copy block number
             let tmp = tweak[8] & 0xf0;
-            tweak[8..].copy_from_slice(&(index as u64).to_be_bytes());
+            tweak[8..].copy_from_slice(&(data_blocks_len as u64).to_be_bytes());
             tweak[8] = (tweak[8] & 0xf) | tmp;
 
             // Last block checksum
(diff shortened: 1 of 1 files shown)


deoxys/src/modes.rs
```diff
@@ -111,75 +111,71 @@ where
...
+            tweak[8] = (tweak[8] & 0xf) | tmp;
 
-                let mut block = Block::default();
-                block[0..data.len()].copy_from_slice(data.get_in());
+            B::encrypt_inout((&mut checksum).into(), tweak.as_ref(), subkeys);
 
-                block[data.len()] = 0x80;
+            for (t, c) in tag.iter_mut().zip(checksum.iter()) {
+                *t ^= c;
+            }
```
newpavlov, line 145: @tarcieri 
I tried to replace this part with `tag ^= checksum;`, but I see that `Array` does not implement the bit op traits. Was it intentional to mirror the built-in arrays?

tarcieri: We could support that, but yeah, core arrays don't, so if we ever try to migrate that back it's one more thing to remove.

FWIW I asked about adding arithmetic op impls to core arrays on IRLO and people did not seem very receptive.

deoxys/src/modes.rs
```diff
@@ -111,75 +111,71 @@ where
...
-            if !data.is_empty() {
-                // Last block, incomplete
+            B::encrypt_inout(data, &tweak, subkeys);
+        }
 
-                // Copy block number
-                let tmp = tweak[8] & 0xf0;
-                tweak[8..].copy_from_slice(&(index as u64).to_be_bytes());
-                tweak[8] = (tweak[8] & 0xf) | tmp;
+        if tail.is_empty() {
```
baloo, line 133: Reversing those branches would make the diff more readable.

deoxys/src/modes.rs
newpavlov: Yeah, I've seen your post and I think I even asked for the same thing somewhere in the rust-lang/rust repo a long time ago with the similar result...

deoxys/src/modes.rs
newpavlov: But it would make the resulting code worse. :)

baloo:
```
        if !tail.is_empty() {
            // Tag computing with incomplete last block
            let index = data_blocks_len;

            // Copy block number
            let tmp = tweak[8] & 0xf0;
            tweak[8..].copy_from_slice(&(index as u64).to_be_bytes());
            tweak[8] = (tweak[8] & 0xf) | tmp;

            // Last block checksum
            tweak[0] = (tweak[0] & 0xf) | TWEAK_M_LAST;

            let mut block = Block::default();
            block[..tail.len()].copy_from_slice(tail.get_in());

            block[tail.len()] = 0x80;

            for (c, d) in checksum.iter_mut().zip(block.iter()) {
                *c ^= d;
            }

            block.fill(0);

            // Last block encryption
            B::encrypt_inout((&mut block).into(), &tweak, subkeys);

            tail.xor_in2out((block[..tail.len()]).into());

            // Tag computing.
            tweak[0] = (tweak[0] & 0xf) | TWEAK_CHKSUM;

            let tmp = tweak[8] & 0xf0;
            tweak[8..].copy_from_slice(&((index + 1) as u64).to_be_bytes());
            tweak[8] = (tweak[8] & 0xf) | tmp;

            B::encrypt_inout((&mut checksum).into(), tweak.as_ref(), subkeys);

            for (t, c) in tag.iter_mut().zip(checksum.iter()) {
                *t ^= c;
            }
        } else {
            // Tag computing without last block
            tweak[0] = (tweak[0] & 0xf) | TWEAK_TAG;

            let tmp = tweak[8] & 0xf0;
            tweak[8..].copy_from_slice(&((buffer_len / 16) as u64).to_be_bytes());
            tweak[8] = (tweak[8] & 0xf) | tmp;

            B::encrypt_inout((&mut checksum).into(), tweak.as_ref(), subkeys);

            for (t, c) in tag.iter_mut().zip(checksum.iter()) {
                *t ^= c;
            }
        }
```

Worse?

baloo: The diff looks like that: https://github.com/RustCrypto/AEADs/compare/master...baloo:baloo/deoxys/modes-tweak?w=1

newpavlov: Ah, you mean the order of blocks in the `tail.is_empty()` part. I erroneously thought that you talked about the `buffer.is_empty()` branch.

The additional `!` makes the code slightly worse in my opinion, but one can argue that the tail block case is more common in practice, so it should go first.

baloo: well, you removed the `buffer.is_empty()` which I completely missed you could do when you initially suggested it. And that made my code messy then.

newpavlov: I found a better way. We do not need the separate branch for the "complete" case in the first place.

PR #694, opened 2025-06-04
belt-dwp: support additional tag sizes

newpavlov:
The specification allows tag truncation (section 5.4). Additionally, DWP can be used with a "full" 128 bit tag size without any security issues (see this [issue](https://github.com/bcrypto/belt/issues/8)). While technically tag sizes bigger than 64 bits do not conform to the spec, I think it's worth to add support for them.

cc @makavity


makavity:
The latest version of the specification was released in 2020 and there are no plans to change it yet. This usually happens after 5 years from the release date.
According to section 5.4, authentication tag is 64 bits, truncation is supported, but not extension.

But i suppose, there is no problem to add this.

PR #703, opened 2025-08-03
implement djb legacy ChaCha20Poly1305 construction

kamulos:
This is a continuation of #304 : I rebased my changes and refactored most of it for better readability. I tried to do it in the most concrete way, without abstractions like the `Cipher` type in the crate. I don't think there is a big chance of code sharing with other algorithms.

Please have a look anytime for feedback. There is definitely some polishing to do, but in general I am happy with the state it is in and I also did some randomized testing with the libsodium. Encryption and decryption between the two libraries seem to work.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -208,6 +208,7 @@ dependencies = [
  "chacha20",
  "cipher",
  "poly1305",
+ "subtle",
  "zeroize",
 ]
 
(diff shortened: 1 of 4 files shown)


tarcieri:
We have a crate for the libsodium-compatible ChaCha20Poly1305 secretbox construction here: https://github.com/RustCrypto/nacl-compat/tree/master/crypto_secretbox

Notably this construction doesn't support AAD, so it isn't actually an AEAD

tarcieri:
TIL there's a *third* construction that calls itself ChaCha20Poly1305. Sigh.

PR #721, opened 2025-09-19
eax: cleanup stale macros

baloo:


newpavlov:
I think you removed a wrong file. It should be `eax/tests/common/mod.rs` instead.

baloo:
Thanks for your attention!

PR #724, opened 2025-09-20
bump aead to latest master, remove heapless

baloo:
Reopening https://github.com/RustCrypto/AEADs/pull/722

I guess I can point to the specific commit of AEAD, before we've split the tests.


newpavlov:
Done as part of #723

PR #739, opened 2025-11-02
aead-stream: remove unnecessary KeyInit constraint on from_aead

rasendubi:
`Encryptor::from_aead` and `Decryptor::from_aead` require AEAD to implement `KeyInit`, which seems unnecessary and unused. I believe we can drop that requirement.


tarcieri (approved):
I guess it's fine, though I can't imagine it being useful in practice

PR #741, opened 2025-11-05
Bump `aead` to v0.6.0-rc.3; restore `getrandom` features

tarcieri:
This release drops the dependency on `rand_core` when using `getrandom`, which was the original feature name for this functionality, and in doing so also avoids a breaking change.

It also updates `rand_core` to the v0.10 prereleases.

Dependencies which depend on `rand_core` v0.10 prereleases have also been bumped to ensure we don't pull in old versions.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5,8 +5,9 @@ version = 4
 
 [[package]]
 name = "aead"
-version = "0.6.0-rc.2"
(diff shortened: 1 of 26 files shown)


tarcieri:
Ended up doing a bunch of documentation cleanup as part of this PR, since much of the documentation is copy-pasted.

Sure could use https://github.com/RustCrypto/traits/pull/1945 for managing the documentation

PR #752, opened 2025-12-26
Bump `aead` dependency to v0.6.0-rc.4

tarcieri:
This is primarily the documentation changes for the newly introduced `Generate` trait which subsumed the previous RNG APIs.

See also:
- RustCrypto/traits#2096
- RustCrypto/traits#2098

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5,9 +5,9 @@ version = 4
 
 [[package]]
 name = "aead"
-version = "0.6.0-rc.3"
+version = "0.6.0-rc.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "03d2d54c4d9e7006f132f615a167865bff927a79ca63d8f637237575ce0a9795"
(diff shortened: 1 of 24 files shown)


tarcieri:
Well, this is odd. Several of the doctests started failing when built without the `getrandom` feature:

```console
running 2 tests
test aes-gcm-siv/src/lib.rs - (line 213) ... ignored
test aes-gcm-siv/src/lib.rs - (line 174) ... FAILED
```

It's not ignoring them in this case for some reason (though curiously, the ones that depend on `arrayvec` and `getrandom` are being ignored).

Also strange is the actual logic doing the feature-based gating didn't change.

tarcieri:
Wow, that's super weird, I can get it to succeed by completely inverting the logic:

```rust
#![cfg_attr(feature = "getrandom", doc = "```ignore")]
#![cfg_attr(not(feature = "getrandom"), doc = "```")]
```

That makes absolutely no sense!

Issue #755, opened 2025-12-26
Doctests no longer build with default features


tarcieri:
This was somehow introduced in #752 which bumped `aead` to v0.6.0-rc.4, but I am baffled as to how or what's even happening.

## Example reproduction

```console
$ cd aes-gcm
$ cargo test --doc
[...]
---- aes-gcm/src/lib.rs - (line 150) stdout ----
error[E0432]: unresolved import `aes_gcm::aead::Generate`
  --> aes-gcm/src/lib.rs:156:28
   |
8  |     aead::{Aead, AeadCore, Generate, Key, KeyInit},
   |                            ^^^^^^^^ no `Generate` in the root
   |
note: found an item that was configured out
```

## Notes

This occurs in several crates, any of which use this style of gating:

```rust
#![cfg_attr(feature = "getrandom", doc = "```")]
#![cfg_attr(not(feature = "getrandom"), doc = "```ignore")]
```

Curiously, if the first line is changed to:

```rust
#![cfg_attr(feature = "getrandom", doc = "```ignore")]
```

Then the test is successfully ignored, suggesting this gating somehow isn't working correctly:

```
running 2 tests
test aes-gcm/src/lib.rs - (line 150) ... ignored
test aes-gcm/src/lib.rs - (line 189) ... ignored
```

Note the other example, which depends on both the `arrayvec` and `getrandom` features, *is* successfully being ignored with logic like:

```rust
#![cfg_attr(all(feature = "getrandom", feature = "arrayvec"), doc = "```")]
#![cfg_attr(
    not(all(feature = "getrandom", feature = "arrayvec")),
    doc = "```ignore"
)]
```

Perhaps the weirdest part is nothing in #752 actually changed how this gating worked, it just started behaving differently when the `aead` crate was upgraded from v0.6.0-rc.3 to v0.6.0-rc.4.

eligrubb:
The issue appears to stem from the `Generate` trait, which the `aead` crate began publicly exporting with `v0.6.0-rc.4` and [is gated by the `rand_core` feature](https://github.com/RustCrypto/traits/blob/b4f864c093cfcb9276e72490aa6fc3e95a4c0c2a/aead/src/lib.rs#L26). This feature is not always simultaneously enabled with the `getrandom` feature, and the rustdocs tests show what happen.

## Potential fixes

### Option A: AEADs level fix

[...]

PR #833, opened 2026-06-17
Make `getrandom` feature imply `rand_core` so Generate doctests compile

eligrubb:
## Issue

Running cargo test (or cargo test --doc) on an individual crate fails to compile the usage doctest:

 ```
   error[E0432]: unresolved import `xaes_256_gcm::aead::Generate`
     --> xaes-256-gcm/src/lib.rs
      |
      |     aead::{Aead, AeadCore, Generate, Key, KeyInit},
      |                            ^^^^^^^^ no `Generate` in the root
      |
   note: found an item that was configured out
      | pub use common::{Generate, rand_core};
      |                  ^^^^^^^^   (gated behind the `rand_core` feature)

   error[E0599]: no associated function named `generate` found ...
 ```

See #755 for more details.

 ## Fix

`Generate` is [exported by `aead` under the `aead/rand_core` feature](https://github.com/RustCrypto/traits/blob/b4f864c093cfcb9276e72490aa6fc3e95a4c0c2a/aead/src/lib.rs#L26). Each AEAD crate currently enables the `aead/rand_core` feature only under their own `rand_core` feature. To fix, chain each crate's `getrandom` feature into its own `rand_core` feature, which transitively enables `aead/rand_core`:

 ```toml
   # before
   getrandom = ["aead/getrandom"]
   # after
   getrandom = ["aead/getrandom", "rand_core"]
 ```

 This matches the `mgm` crate which already follows this convention.

Crates touched:
 - aes-gcm, aes-gcm-siv, aes-siv, ascon-aead128, belt-dwp, ccm, chacha20poly1305, deoxys, eax, ocb3: ["aead/getrandom"] → ["aead/getrandom", "rand_core"]
 - xaes-256-gcm: ["aes-gcm/getrandom"] → ["aead/getrandom", "aes-gcm/getrandom", "rand_core"]

 ## Testing

 - Each crate's doctests now pass standalone: cargo test --doc -p <crate> for all 11.
 - cargo test --workspace still passes

## Note

Alternative, simpler upstream fix may exist, all `getrandom` features already enable `aead/getrandom`. If upstream, `aead` feature `aead/getrandom` implied `aead/rand_core`, this would not be necessary. See [`traits` PR#2450](https://github.com/RustCrypto/traits/pull/2450) for a proposal of how this would look.


eligrubb:
Closing in favor of [traits PR#2452]() which solves this problem upstream. `aead` dependency version will need to be bumped before tests are fixed and #755 can be closed.

PR #863, opened 2026-07-22
xaes-256-gcm: add `zeroize` feature (#863)

### LINKED ISSUE #862: xaes-256-gcm: `zeroize` feature is not implemented for crate
Unlike other crates in `AEADs`, and other RustCrypto respositories, the `xaes-256-gcm` crate does not offer an optional `zeroize` feature or implement the `ZeroizeOnDrop` trait. While zeroization is generally a defense in depth attribute and not a security guarantee, offering zeroization in the `xaes-256-gcm` crate would decrease attack surface area for recovering key data.

The majority of sensitive data handling in `xaes-256-gcm` is via the underlying `aes256gcm` operation. `aes`, `aes_gcm`, and `cipher`, all currently implement the `zeroize` feature and could be used to add the feature to `xaes-256-gcm`.

eligrubb:
Fixes #862, adding optional `zeroize` feature to `xaes-256-gcm` crate.

Follows pattern of other crates, using marker trait `ZeroizeOnDrop` to cascade trait to underlying `Aes256`. Zeroizes `k1` block on drop. When feature is enabled, zeroizes local variable `k` that contains raw key material.

As part of this PR, the helper method `derive_key`'s signature was changed to avoid cloning sensitive key material on the stack. Instead, the helper now accepts a pointer to the key material and modifies it in place. This avoids the helper method copying key material into its local stack, which is not zeroized when `derive_key` returns.

### DIFF
--- a/xaes-256-gcm/src/lib.rs
+++ b/xaes-256-gcm/src/lib.rs
@@ -118,8 +118,17 @@ impl AeadInOut for Xaes256Gcm {
         }
 
         let (n1, n) = nonce.split_ref::<<NonceSize as Div<U2>>::Output>();
-        let k = self.derive_key(n1);
-        Aes256Gcm::new(&k).encrypt_inout_detached(n, associated_data, buffer)
+        let mut k = Key::<Aes256Gcm>::default();
+        self.derive_key(n1, &mut k);
+        let cipher = Aes256Gcm::new(&k);
+
+        #[cfg(feature = "zeroize")]
+        {
+            use zeroize::Zeroize;
+            k.as_mut_slice().zeroize();
+        }
+
+        cipher.encrypt_inout_detached(n, associated_data, buffer)
     }
 
     fn decrypt_inout_detached(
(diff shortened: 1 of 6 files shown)


xaes-256-gcm/src/lib.rs
```diff
@@ -158,7 +176,6 @@ impl Xaes256Gcm {
         // Kₘ = AES-256ₖ(M1 ⊕ K1)
         // Kₙ = AES-256ₖ(M2 ⊕ K1)
         // Kₓ = Kₘ || Kₙ = AES-256ₖ(M1 ⊕ K1) || AES-256ₖ(M2 ⊕ K1)
-        let mut key: Key<Aes256Gcm> = Array::default();
```
eligrubb, line 161: As it was, this helper method would allocate `key` on the local stack, and then copy raw key material into it. Then, it would copy the result into the return buffer, never clearing the stack-allocated key material. This PR instead allocated the key buffer first, before calling `derive_key`, and passes in a reference.

tarcieri (approved):

PR #885, opened 2026-09-05
fix(aes-gcm): reject empty nonces

thevilledev:
## Summary

Running the [Wycheproof AES-GCM vectors](https://github.com/C2SP/wycheproof/blob/main/testvectors_v1/aes_gcm_test.json) exposed failures for the `ZeroLengthIv` cases. A zero-length nonce violates [NIST SP 800-38D](https://csrc.nist.gov/pubs/sp/800/38/d/final) and can disclose the GHASH authentication key.

## Changes

- Add an inline const assertion requiring non-zero nonce sizes.
- Enforce the assertion in the counter initialization shared by encryption and decryption.
- Add a compile-fail regression for `AesGcm<Aes128, U0>`.

## Tests

Tests pass.


tarcieri:
This is probably a good idea in general but it's also imposing a new bound on a generic parameter, which is a breaking change: anyone using it in a generic context will need to propagate this bound.

newpavlov:
As a temporary solution we could panic on empty nonces. I doubt that anyone uses the crate with them, so it should not affect anyone while serving as a guard against misuse.

tarcieri:
Hmm, would a const assert work, or would it complain about the outer generic parameter?

thevilledev:
> Hmm, would a const assert work, or would it complain about the outer generic parameter?

Good idea. I'll think the only caveat is that `cargo check` doesn't detect it? I'll try this out

